PR #2, opened 2012-08-27
[MENFORCER-138] Rule to ban all transitive dependencies

jsenko:
In some projects it's desirable to have all dependencies specified in pom. It would be nice to have an enforcer rule that would ban all transitive dependencies so that the user could either add the transitive dependency directly to the pom (if it's actually needed), or exclude the dependency.
http://jira.codehaus.org/browse/MENFORCER-138

This rule will ban all transitive dependencies, unless configured otherwise.
The configuration works identically to http://maven.apache.org/enforcer/enforcer-rules/bannedDependencies.html .
User can exclude certain dependencies from being checked (anywhere in the dependency tree) add also add exceptions to the excludes for finer configuration (see javadoc for details).

If it fails, the rule will generate message showing dependencies which caused the failure and their descendants in a tree, with info which dependencies were excluded in that branch. Alternatively, custom message can be used.

### DIFF
--- a/enforcer-rules/pom.xml
+++ b/enforcer-rules/pom.xml
@@ -87,7 +87,7 @@
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-dependency-tree</artifactId>
-      <version>1.2</version>
+      <version>2.0</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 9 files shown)


jira-importer:
Resolve #411

PR #17, opened 2016-03-14
Added support for reading banned exclusions from a URL

subashiniunny:


michael-o:
We aren't going to accepted externally patched versions of our components. Please redo or close this PR.

PR #18, opened 2016-04-25
[MENFORCER-247] Add a "require file checksum" rule

lyubomyr-shaydariv:
https://issues.apache.org/jira/browse/MENFORCER-247

### DIFF
--- a/enforcer-rules/pom.xml
+++ b/enforcer-rules/pom.xml
@@ -67,6 +67,10 @@
       <artifactId>commons-lang</artifactId>
     </dependency>
     <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.maven.enforcer</groupId>
       <artifactId>enforcer-api</artifactId>
     </dependency>
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireFileChecksum.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireFileChecksum.java
@@ -112,9 +112,9 @@ public class RequireFileChecksum
         InputStream inputStream = null;
         try
         {
-            if ( !this.file.canRead() )
+            if ( this.file.isDirectory() || !this.file.canRead() )
             {
-                throw new EnforcerRuleException( "Missing file: " + this.file );
+                throw new EnforcerRuleException( "Cannot read file: " + this.file.getAbsolutePath() );
             }
 
             inputStream = new FileInputStream( this.file );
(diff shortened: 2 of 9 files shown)


khmarbaise:
This looks good so far. I will take a look within the next few days...

Tunaki:
Some comments: an enumeration of the different types could be used. It is possible to inject directly an enum value: it is mapped by its `name()` (so the configuration in the POM would be `"MD5"` for example, for the enum value `MD5`).

Also, the tests could use the JUnit rule `TemporaryFolder` instead of creating and handling temporary files manually. It would automatically take care of deleting them, whatever the result of the tests.

Last point, it would be nice to add an IT as well inside {{maven-enforcer-plugin}}. It would test the rule end-to-end.

lyubomyr-shaydariv:
I've just pushed a f ew commits including the null-args fix and the _is-a-directory_ check.

@Tunaki 
Thanks for pointing the `TemporaryFolder` rule. However, I'm not sure about items 1 and 3:

* I would not use an enum `.name()` value here because strings are somewhat more flexible. I'm just trying not to add new classes to a project where I'm just a passing-by contributor, or is there an appropriate enum around already? However, it might be a good point to give the hash types some better names if the current ones are not good (`md5`, `sha1`, `sha256`, `sha384`, and `sha512`) -- maybe something like `MD5`, `SHA-1`, `SHA-256`, `SHA-384`, and `SHA-512` (cannot be implemented using `Enum.name()`)?
* Regarding the IT tests, I'd appreciate if I could be given an example from the existing codebase.

Tunaki:
No, there is no enum already, but it could be added as an nested class. That can be done later though, if the need arises, so I agree with keeping the current strings.

For the IT, yes, you can look at the [`require-files-size`](https://github.com/apache/maven-enforcer/tree/trunk/maven-enforcer-plugin/src/it/projects/require-files-size) one for example. I imagine a new `require-file-checksum` folder in `src/it/projects` that would calculate the checksum of a sample file and verify it against pre-calculated values.

lyubomyr-shaydariv:
Added an integration test (the code borrowed from a nearby `require-files-exist` test). I'm not sure if I did it right since I had to replace `@project.version@` with `1.4.2-SNAPSHOT` manually in order to verify it. Maybe I missed something else?

Anyway, I shall be hard to reach the next two days, so have a nice weekend!

Tunaki:
Ah yes, I forgot to say that the ITs are ran with by launching the build with the profile `run-its`. A full build of the Enforcer Plugin, and running the ITs, would be done with the command `mvn clean install -Prun-its`.

You don't need to set the version manually. `@project.version@` is automatically replaced when the test is invoked by the version of the project being built. It makes sure that the tests are always ran on the current version.

lyubomyr-shaydariv:
@Tunaki 
thanks for the tip. Running the ITs with that profile enabled failed due to a missing ASF license header, so I just replaced the old dummy file with `LICENSE` containing just the header. Not sure if it's fine now, though.

Tunaki:
This is fine, yes. Thanks for the work! I'll get around to merging that.

jira-importer:
Resolve #750

jira-importer:
Resolve #750

PR #20, opened 2016-11-14
MENFORCER-195 - use DependencyGraphBuilder API instead of DependencyTreeBuilder

ctomc:
Using this I get enforcer to work with wildcard exclusion without any issue.
as a bonus ban bad dependencies rule also gets much faster (0.5sec vs 15sec) in our case.
Initial issue that was mentioned in MENFORCER-195 was somewhat fixed as part of work on MDEP-374 so it is safe to use new API.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/DependencyConvergence.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/DependencyConvergence.java
@@ -24,10 +24,6 @@ import java.util.Collections;
 import java.util.List;
 
 import org.apache.maven.artifact.Artifact;
-import org.apache.maven.artifact.factory.ArtifactFactory;
-import org.apache.maven.artifact.metadata.ArtifactMetadataSource;
-import org.apache.maven.artifact.repository.ArtifactRepository;
-import org.apache.maven.artifact.resolver.ArtifactCollector;
 import org.apache.maven.artifact.resolver.filter.ArtifactFilter;
 import org.apache.maven.enforcer.rule.api.EnforcerRule;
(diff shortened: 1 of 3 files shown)


Tunaki:
The `DependencyGraphBuilder` component still does not handle omitted nodes. This is tracked in [MSHARED-339](https://issues.apache.org/jira/browse/MSHARED-339) and is not resolved yet. This change is thus breaking some ITs for now.

ctomc:
From what I tried it looks like the fix for https://issues.apache.org/jira/browse/MDEP-374 handled that fine.
But to be fair, I did test only with latest released versions of maven and dependency plugin.

Also tests on our codebase showed that duplicate deps ware found as expected.

Can you point me to ITs this breaking?

Tunaki:
Yes, you can build the `maven-enforcer` project and run its ITs with `mvn clean verify -Prun-its`. I ran them with Maven 3.3.9 and there were 6 failing ITs (after fixing small checkstyle issues):

```
[ERROR] *  dependencies_not_converging\pom.xml
[ERROR] *  dependencies_not_converging_test_scope\pom.xml
[ERROR] *  require-upper-bound-dependencies-managed_failure\pom.xml
[ERROR] *  require-upper-bound-dependencies-unique_failure\pom.xml
[ERROR] *  require-upper-bound-dependencies-unique_success\pom.xml
[ERROR] *  require-upper-bound-deps_failure\pom.xml
```

Version 3.0 of the `maven-dependency-tree` shared component is Maven 3 specific and the plugin isn't migrated to it yet, so we stick to 2.2. But even with 3.0, I don't think it'll solve the issue, the Dependency Plugin also doesn't support `verbose` currently (which would show omitted nodes).

jira-importer:
Resolve #692

PR #21, opened 2017-01-17
Add rule: banProfiles

mikkoi:
Same patch as in https://issues.apache.org/jira/browse/MENFORCER-261

Signed-off-by: Mikko Johannes Koivunalho <mikko.koivunalho@iki.fi>


jira-importer:
Resolve #434

PR #25, opened 2017-07-04
MENFORCER-276 - Support ignoring dependency scopes in RequireUpperBoundDeps

oleg-nenashev:
We have recently adopted `RequireUpperBoundDeps` in the Jenkins project (https://github.com/jenkinsci/plugin-pom/pull/67). In order to implement it, Jesse Glick added a support of ignoring particular dependencies in [MENFORCER-273](https://issues.apache.org/jira/browse/MENFORCER-273).

Sometimes we declare dependencies between plugins for testing purposes (scope=test), and in such case RequireUpperBoundDeps fails the build if there is a test dependency conflicting with the main code. Although it may cause instability of tests (including false positive results), it does not really impact the distributable package in our case.

I propose to add another option for filtering of dependency scopes. This pull request also adds some unit tests for `RequireUpperBoundDeps`. Previously it had integration tests only.

https://issues.apache.org/jira/browse/MENFORCER-276

@reviewbybees @jglick @stephenc

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
@@ -71,7 +71,7 @@ public class RequireUpperBoundDeps
     private List<String> excludes = null;
     
     /**
-     * If {@code true}, upper bounds from test dependencies will be ignored.
+     * Dependency scopes to ignore.
      * 
      * @since TBD
      */
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
@@ -21,9 +21,11 @@ package org.apache.maven.plugins.enforcer;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.factory.ArtifactFactory;
@@ -67,6 +69,13 @@ public class RequireUpperBoundDeps
      * @since TBD
      */
     private List<String> excludes = null;
+    
+    /**
+     * If {@code true}, upper bounds from test dependencies will be ignored.
+     * 
+     * @since TBD
+     */
+    private List<String> ignoredDependencyScopes;
 
     /**
      * Set to {@code true} if timestamped snapshots should be used.
@@ -86,6 +95,17 @@ public class RequireUpperBoundDeps
     public void setExcludes( List<String> excludes )
     {
         this.excludes = excludes;
+    }
+
+    /**
+     * Ignores particular dependency scopes.
+     * 
+     * @param scopes a list of ignored scopes
+     * @since TBD
+     */
+    public void setIgnoreDependencyScopes( List<String> scopes ) 
+    {
+        this.ignoredDependencyScopes = scopes;
     }
 
     // CHECKSTYLE_OFF: LineLength
@@ -143,8 +163,8 @@ public class RequireUpperBoundDeps
         try
         {
             DependencyNode node = getNode( helper );
-            RequireUpperBoundDepsVisitor visitor = new RequireUpperBoundDepsVisitor();
-            visitor.setUniqueVersions( uniqueVersions );
+            RequireUpperBoundDepsVisitor visitor = new RequireUpperBoundDepsVisitor
+                    ( uniqueVersions, ignoredDependencyScopes, log );
             node.accept( visitor );
             List<String> errorMessages = buildErrorMessages( visitor.getConflicts() );
             if ( errorMessages.size() > 0 )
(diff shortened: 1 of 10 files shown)


enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
```diff
@@ -67,6 +69,13 @@
      * @since TBD
      */
     private List<String> excludes = null;
+    
+    /**
+     * If {@code true}, upper bounds from test dependencies will be ignored.
```
jglick: Well, from the specified scopes.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
```diff
@@ -242,18 +262,33 @@ private String getFullArtifactName( DependencyNode node, boolean usePremanaged )
...
         }
 
         private Map<String, List<DependencyNodeHopCountPair>> keyToPairsMap =
             new LinkedHashMap<String, List<DependencyNodeHopCountPair>>();
 
         public boolean visit( DependencyNode node )
         {
+            Artifact artifact = node.getArtifact();
+            String artifactScope = artifact.getScope();
+            if ( ignoredDependencyScopes != null && ignoredDependencyScopes.contains( artifactScope ) )
```
jglick: I think this filtering should rather be done in `buildErrorMessages`. It is unclear to me what the effect will be of filtering during visits to (potentially transitive) dependencies but it will probably be wrong, since the scope declared at top level is not necessarily the same as that used throughout the dependency chain.

maven-enforcer-plugin/src/it/projects/require-upper-bound-test-deps_failure/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  * KIND, either express or implied.  See the License for the
+  * specific language governing permissions and limitations
+  * under the License. 
+  *
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>org.apache.maven.plugins.enforcer.its</groupId>
+  <artifactId>menforcer128</artifactId>
```
jglick: Copy-paste error?

maven-enforcer-plugin/src/it/projects/require-upper-bound-test-deps_failure/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  <version>1.0-SNAPSHOT</version>
+  <packaging>jar</packaging>
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven.plugins.enforcer.its</groupId>
+      <artifactId>menforcer128_api</artifactId>
```
jglick: ditto

maven-enforcer-plugin/src/it/projects/require-upper-bound-test-deps_failure/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+      <version>1.4.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven.plugins.enforcer.its</groupId>
+      <artifactId>menforcer128_classic</artifactId>
```
jglick: ditto

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
oleg-nenashev: Yeah. I started from boolean flag and forgot to fix Javadoc

maven-enforcer-plugin/src/it/projects/require-upper-bound-test-deps_failure/pom.xml
oleg-nenashev: Yeah, I didn't change it

oleg-nenashev:
> Something like this would be a reasonable enhancement for the enforcer rule generally, but I would not advocate using it in the Jenkins parent POM. In fact enforcing upper bounds on test dependencies was very much a goal of using this rule to begin with.

Yes, in Jenkins context I consider it only as a opt-in feature for plugins. The default aggressive behavior looks reasonable to me.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
oleg-nenashev: The code is correct for the `test` scope, but I agree that it may be different for other scopes

maven-enforcer-plugin/src/it/projects/require-upper-bound-test-deps_failure/pom.xml
oleg-nenashev: Well, I copy-pasted it from tests for MENFORCER-273, which also do not have it updated :)

elharo:
Since this is several years old, I'm going to close it. Feel free to open a new PR that resolves the merge conflicts.

uhafner:
@oleg-nenashev Don't we need this PR anymore in Jenkins? I am currently struggling with a lot of enforcer failures in my test dependencies so having such a feature would be nice.

uhafner:
Example: 
```
[WARNING] Rule 0: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for net.bytebuddy:byte-buddy:1.10.10 paths to dependency are:
+-io.jenkins.plugins:warnings-ng:8.5.0-SNAPSHOT
  +-org.xmlunit:xmlunit-assertj:2.8.1
    +-net.bytebuddy:byte-buddy:1.10.10
and
+-io.jenkins.plugins:warnings-ng:8.5.0-SNAPSHOT
  +-org.mockito:mockito-core:3.6.0
    +-net.bytebuddy:byte-buddy:1.10.15
]

```

aheritier:
I got exactly the same issue yesterday on a non-jerkins project too. I agree that we should try to have it fixed.

oleg-nenashev:
I am still around if any help is needed

jglick:
@uhafner the error message in your case looks legitimate. https://github.com/apache/maven-enforcer/pull/25#pullrequestreview-47866562

jira-importer:
Resolve #763

jira-importer:
Resolve #763

jira-importer:
Resolve #763

PR #29, opened 2017-11-05
Update to Mockito 2.11

slachiewicz:
Fixes MENFORCER-278


slachiewicz:
Fixed with merge for MENFORCER-278

PR #32, opened 2018-03-05
[MENFORCER-298] Ensure bannedDependencies does not corrupt the classpath when using timestamped snapshots

jglick:
See background in [MENFORCER-298](https://issues.apache.org/jira/browse/MENFORCER-298).

This patch is confirmed to fix the problem on the originally discovered test case. It does not purport to fix problems when using

```xml
<searchTransitive>false</searchTransitive>
```

(for which no ITs seem to exist), nor with other calls to `buildDependencyGraph`, notably from `banTransitiveDependencies` or `banCircularDependencies` in `extra-enforcer-rules`. Those _might_ be fixable using `MavenProject.getDependencyArtifacts` and/or `Artifact.getDependencyTrail`.

I attempted to write an IT to reproduce the bug but failed: `mrm-maven-plugin` refused to serve the snapshot JAR artifacts needed for my test even though I tried placing them, along with POMs and `maven-metadata.xml`, in `maven-enforcer-plugin/src/it/mrm/repository/` using

* a flat directory structure, as in current ITs in this repo
* an unpacked JAR format-2 layout as the [mrm documentation seems to suggest](http://www.mojohaus.org/mrm/mrm-maven-plugin/examples/invoker-tests.html)
* same, but with prepacked JARs, as `maven-integration-testing` does

I just copied the basic test setup from `MavenITmng4189UniqueVersionSnapshotTest` with minor modifications. If anyone is interested I can file a PR showing my attempt.

@stephenc suggested that my simple call to `MavenProject.getArtifacts` might have some undesirable side effects relating to overeager download of dependencies. Whatever the case, my patch does not regress any of the existing unit or integration tests.

@reviewbybees for my colleagues

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractBanDependencies.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractBanDependencies.java
@@ -63,22 +63,25 @@ public abstract class AbstractBanDependencies
             throw new EnforcerRuleException( "Unable to retrieve the MavenProject: ", eee );
         }
 
-        try
+        if ( ! searchTransitive )
         {
-            graphBuilder = (DependencyGraphBuilder) helper.getComponent( DependencyGraphBuilder.class );
-        }
-        catch ( ComponentLookupException e )
-        {
-            // real cause is probably that one of the Maven3 graph builder could not be initiated and fails with a
-            // ClassNotFoundException
             try
             {
-                graphBuilder =
-                    (DependencyGraphBuilder) helper.getComponent( DependencyGraphBuilder.class.getName(), "maven2" );
+                graphBuilder = (DependencyGraphBuilder) helper.getComponent( DependencyGraphBuilder.class );
             }
-            catch ( ComponentLookupException e1 )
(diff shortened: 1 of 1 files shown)


enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractBanDependencies.java
```diff
@@ -63,22 +63,25 @@ public void execute( EnforcerRuleHelper helper )
             throw new EnforcerRuleException( "Unable to retrieve the MavenProject: ", eee );
         }
 
-        try
```
jglick: Note that [most of the changes are just reindentation](https://github.com/apache/maven-enforcer/pull/32/files?w=1).

elharo:
Since this is several years old, I'm going to close it. Feel free to open a new PR that resolves the merge conflicts.

jglick:
Sigh. My project is no longer using timestamped snapshots in most cases so I do not care to spend the ten minutes recreating this. If someone else is affected by the bug I guess they can try it, and reverify behavior in their situation.

jira-importer:
Resolve #779

PR #33, opened 2018-04-16
[MENFORCER-301] check dependencyManagement

kudrevatykh:
really check dependencyManagement section

### DIFF
--- a/maven-enforcer-plugin/src/it/projects/ban-duplicate-dependencies-versions/pom.xml
+++ b/maven-enforcer-plugin/src/it/projects/ban-duplicate-dependencies-versions/pom.xml
@@ -1,0 +1,69 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


kudrevatykh:
added tests and rebased on master

maven-enforcer-plugin/src/it/projects/ban-duplicate-dependencies-versions/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project>
+    <modelVersion>4.0.0</modelVersion>
```
elharo, line 23: nit: maven style for XML files is two space indents:

kudrevatykh: fixed indent

pzygielo:
I see for new IT that build fails with
```
Found 1 duplicate dependency declarations in this project
```
as expected.
But could it be **verified**, to exclude possible future failures due to **other** than new rule violation reasons? (For example adding _unrecognised tag_ in pom.xml also results in IT build failure which is incorrectly recognized as passed test.)

kudrevatykh:
> I see for new IT that build fails with
> 
> ```
> Found 1 duplicate dependency declarations in this project
> ```
> 
> as expected.
> But could it be **verified**, to exclude possible future failures due to **other** than new rule violation reasons? (For example adding _unrecognised tag_ in pom.xml also results in IT build failure which is incorrectly recognized as passed test.)

added verification script

jira-importer:
Resolve #778

PR #48, opened 2018-12-11
[MENFORCER-324] - Shouldn't we use System.lineSeparator() instead?

jsoref:

       pulling in other changes.
       commit message.

This tries to address a request by @khmarbaise in https://github.com/apache/maven-enforcer/pull/47#discussion_r240368459

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractBanDependencies.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractBanDependencies.java
@@ -95,7 +95,7 @@ public abstract class AbstractBanDependencies
             StringBuilder buf = new StringBuilder();
             if ( message != null )
             {
-                buf.append( message + "\n" );
+                buf.append( message + System.lineSeparator() );
             }
             for ( Artifact artifact : foundExcludes )
             {
@@ -110,7 +110,7 @@ public abstract class AbstractBanDependencies
 
     protected CharSequence getErrorMessage( Artifact artifact )
     {
-        return "Found Banned Dependency: " + artifact.getId() + "\n";
+        return "Found Banned Dependency: " + artifact.getId() + System.lineSeparator();
     }
 
     protected Set<Artifact> getDependenciesToCheck( MavenProject project )
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractRequireFiles.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/AbstractRequireFiles.java
@@ -90,7 +90,7 @@ public abstract class AbstractRequireFiles
             StringBuilder buf = new StringBuilder();
             if ( message != null )
             {
-                buf.append( message + "\n" );
+                buf.append( message + System.lineSeparator() );
             }
             buf.append( getErrorMsg() );
 
@@ -98,11 +98,11 @@ public abstract class AbstractRequireFiles
             {
                 if ( file != null )
                 {
-                    buf.append( file.getAbsolutePath() + "\n" );
+                    buf.append( file.getAbsolutePath() + System.lineSeparator() );
                 }
                 else
                 {
-                    buf.append( "(an empty filename was given and allowNulls is false)\n" );
+                    buf.append( "(an empty filename was given and allowNulls is false)" + System.lineSeparator() );
                 }
             }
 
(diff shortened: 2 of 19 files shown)


eolivelli:
@jsoref thank you for sending this patch.

Before merging this PR I have to understand @khmarbaise  and @Tibor17 concerns.
Does any Windows user ever reported problems with new lines ? I guess that Java or some other layer translates '\n' to the correct EOL depending on the platform.

jsoref:
@eolivelli: I'm not endorsing this PR. I provided it based on their requests. I'm using macOS and this isn't a concern I have.

If you're curious about the behavior, you can use

Microsoft [Win 7/Win 8.1/Win 10](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/) in [VirtualBox](https://www.virtualbox.org/wiki/Downloads) for free and install Java/maven/maven-enforcer.

Tibor17:
@eolivelli 
If the user does not want to continue, we can take over.
Java is quite redundant language and we can make better if we put a public static final constant `NL` in `EnforcerUtils` and replace all these calls. These changes remind me to say that there are these code constructs:
`stingBuilder.append( "a" + "b" )` which should be transformed to two calls of `append`.
 If we do it, we should do it properly.
On the other side the reason for these strings is displaying a message on console and the messages are never so big that we have to use `StringBuilder` as performance optimizer. Humans are not able to react on changes below 0.3 seconds and these CPUs can concatenate Kilo bytes of strings within microseconds.
There I use to make a compromises in my codes because `javac 1.5+` translates simple concatenation `"a" + "b"` to the same code with `StringBuilder` in bytecode except the loops. Making more compact Java code with a slowness cca 1 microsecond is still perfect for creating these Exception messages. And anyway if we construct Exception it means that the build is going to fail and there such delay 1E-6 seconds does not matter at all even if it was 1E-3 seconds.

eolivelli:
I think that we don't need this change if noone ever reported errors on Windows and Mac.

I appreciate @jsoref work and the fact that he is responding to suggestions and requests of the maintainers of this code base.

The initial request was Karl

Tibor17:
@eolivelli 
Correct!
I think the main change in https://github.com/apache/maven-enforcer/pull/47 can be pushed even with `\n`. If somebody is shouting in users, we will fix new lies then.

jsoref:
@eolivelli: rebased, but if you guys aren't particularly certain you want to take it, I really don't want to push this much past where it is now.

And yes, I'm aware of @Tibor17 's concerns w/ `StringBuilder`, I considered it and concluded for more or less the same reasons that @Tibor17 did that there wasn't any particular value in doing more than I did. -- I used a set of pure search and replaces (plus eventually fixing line widths), whereas proper StringBuilder would have required me to actually read all the code.

michael-o:
@eolivelli @Tibor17 anything holding us off merging @jsoref's work? Looks good to me now.

Tibor17:
No objections from me. Go ahead!

eolivelli:
Seems to me and to the contributor himself that this change is not needed.
Are there any other cases in other plugins ?
We should keep consistent behavior among all the plugins.

I am -0 on this patch.
Feel free to commit but I think we can leave without it.

I appreciate the work and patience of @jsoref . Thank you

michael-o:
@eolivelli why? Line ending consistency is important. There a hundreds of spots in our codebase which is inconsistent. I, therefore, appreciate this PR.

eolivelli:
@michael-o does any user ever reported any problem with the output of this plugin ?

I think that somehow on Windows (on Linux I am sure it is already working well) there is something which is making the output nicer. Something that is already translating the new line to the corresponding platform dependant value.

I think that that constant is to be used when writing to a file for instance, not to stdout.

So if no one ever complained why changing the code ? We could introduce some regression hard to  debug.
What's your opinion ?

michael-o:
No one needs to complain to make things better. I fixed a lot of stuff last year in Maven SCM where no one complained, but I have stumbled upon because they were just broken in that specific constellation. Having a consistent `stdout` which can be written to a file is a good thing.

eolivelli:
@michael-o 
fine to me, let's merge this one.
Thank you for your clarification.

I feel it would be useful to raise this point to dev list and clean up other plugins. So that when a committer works on a plugin he has a change to clean up line endings writes to stdout

michael-o:
@eolivelli I have done that actually in places I have found. Feel free to work on that kind of stuff and request a review by some fellow committer.

You should always consider compared to our huge user base we receive a very small amount of usable contributions.

Please go on, triage that ticket and merge the commit with the ticket id.

@jsoref thanks again!

eolivelli:
Okay so I will continue with this ticket and eventually merge it to master.

eolivelli:
Last run of CI, then merging
https://builds.apache.org/job/maven-box/job/maven-enforcer/job/MENFORCER-324/

jira-importer:
Resolve #803

PR #58, opened 2019-12-14
[MENFORCER-288] Support new Java versioning schema for older versions.

bmhm:
Signed-off-by: Benjamin Marwell <bmarwell@gmail.com>

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MENFORCER) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MENFORCER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 

 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVersion.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVersion.java
@@ -22,6 +22,7 @@ package org.apache.maven.plugins.enforcer;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
@@ -39,6 +40,12 @@ import org.codehaus.plexus.util.StringUtils;
 public class RequireJavaVersion
     extends AbstractVersionEnforcer
 {
+
+    /**
+     * Pattern which matches Java versions 6, 7, 8 (and earlier).
+     */
+    private static final Pattern PATTERN_MAIN_OLD_JDK = Pattern.compile( "^[1-8]$" );
+
     @Override
     public void execute( EnforcerRuleHelper helper )
         throws EnforcerRuleException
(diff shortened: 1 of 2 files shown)


bmhm:
While I still not like the forced else, I added it as requested in other PRs. Note that this will make the commit much bigger, because there is need to re-indent a lot of code. If granted, I will happily change it to:

```java
        if ( PATTERN_MAIN_OLD_JDK.matcher( theJdkVersion ).matches() )
        {
            return "1." + theJdkVersion;
        }
```

This would remove all those unnecessary changes:
https://github.com/apache/maven-enforcer/pull/58/files#diff-e944db4bfa78de9044a8e0b7ff8f51c5R81-R112

rfscholte:
Consider using https://codehaus-plexus.github.io/plexus-languages/plexus-java/apidocs/org/codehaus/plexus/languages/java/version/JavaVersion.html whioch is already used by several other maven-plugins

bmhm:
> Consider using https://codehaus-plexus.github.io/plexus-languages/plexus-java/apidocs/org/codehaus/plexus/languages/java/version/JavaVersion.html whioch is already used by several other maven-plugins

The contribution docs state that changes should be kept minimal. Please open an issue "migrate to `JavaVersion`", then I'll be happy to submit another PR.

bmhm:
@rfscholte 
I just did take a look at `JavaVersion` from plexus-java you linked. That would not be much help, because even if you input "8" it cannot convert to `1.8`. I'd have to rewrite the whole logic. I'm not doing this without consent.

bmhm:
Oh I think I missed the most important part:
```java
        enforceVersion( helper.getLog(), "JDK", getVersion(), detectedJdkVersion );
```
should become
```java
        enforceVersion( helper.getLog(), "JDK", normalizeJDKVersion( getVersion() ), detectedJdkVersion );
```

rfscholte:
The rule `requireJavaVersion` is about comparing version, which is the main task of `JavaVersion`.
I'm pretty sure the `normalizeJDKVersion` can be replaced/removed. (and it would make comparison equal to other plugins)

bmhm:
> I'm pretty sure the normalizeJDKVersion can be replaced/removed. (and it would make comparison equal to other plugins)

Yes, but now I'm building a method to parse `getVersion()` which could be anything like "[8,11]" or "(,1.7],[1.8,),6,7" or something. They need to get normalized first, don't they? That's a LOT of  probable impact. And a lot of tests cannot be used anymore. :(

bmhm:
Actually, `enforceVersion` compares the systemProperty against a version range. The version range must be normalized or each put through JavaVersion.parse(). But this will not help, because if the range says (6 through 8), JavaVersion.compare will not accept with a VersionRange argument.

Example:
I use 
```java
<requireJavaVersion>
    <version>(,7),(7,)</version> <!-- not java 7 -->
</requireJavaVersion>
```

Then I get `JavaVersion("1.7")` from the system property. The `compareTo` method is useless, and the version range cannot be normalized. I do not thing `getVersion` should be parsed before handling over to the `enforceVersion` method. Should it?

bmhm:
Thinking about what I just wrote:
Accepting a version range "6" but not "[6,7]" does not make sense. I'll remove this PR.

jira-importer:
Resolve #772

PR #59, opened 2019-12-24
[MENFORCER-336] - Fixes race condition in EvaluateBeanshell

vbreivik:
The Beanshell interpreter in EvaluateBeanShell was a static field, but this object is not thread safe. I wrapped the BeanShell interpreter in a ThreadLocal. The alternative would be to create a new Interpreter on each invocation, but I don't know how expensive the Interpreter object is. Using ThreadLocal is probably closer to the original design where it was a singleton. 

I added a (somewhat naive) unit test that failed on every run on my old 4 core mac, and ran green with the new ThreadLocal variant. However I am not sure if this is the correct way to test this, or how this will behave on other hardware. 

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MENFORCER-336) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MENFORCER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 

 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/EvaluateBeanshell.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/EvaluateBeanshell.java
@@ -38,7 +38,14 @@ public class EvaluateBeanshell
 {
 
     /** Beanshell interpreter. */
-    private static final Interpreter BSH = new Interpreter();
+    private static final ThreadLocal<Interpreter> INTERPRETER = new ThreadLocal<Interpreter>()
+    {
+        @Override
+        protected Interpreter initialValue()
+        {
+            return new Interpreter();
+        }
+    };
 
     /** The condition to be evaluated.
      *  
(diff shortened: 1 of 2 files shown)


vbreivik:
> Thanks. I think this change might require signing an Apache Individual Contributor License Agreement.

Ah, sorry about that. I won't have access to a printer and scanner for some weeks, but I will sign and send it when I get a chance.

olamy:
the change is not that big. we don't need a CLA. by contributing you agree to the license it's implicit and by the way you do not add any new files but simply modify files under Apache license.

vbreivik:
> the change is not that big. we don't need a CLA. by contributing you agree to the license it's implicit and by the way you do not add any new files but simply modify files under Apache license.

I will leave it up to you to decide. I do have another pull request open for another plugin, so I will sign the CLA when I get access to a scanner again.

jira-importer:
Resolve #812

PR #78, opened 2020-09-09
[MENFORCER-360] - requireUpperBoundDeps should have option to check for same major version

guyv:
An option is added to the built-in requireUpperBoundDeps rule to check for same major version (following semver.org).

So a (transitive) dependency on groupId:artifactId:1.0.0 and on groupId:artifactId:2.0.0 means that we have a conflict.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireUpperBoundDeps.java
@@ -62,6 +62,11 @@ public class RequireUpperBoundDeps
     private boolean uniqueVersions;
 
     /**
+     * @since 3.0.0
+     */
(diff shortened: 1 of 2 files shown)


elharo (commented):
@rfscholte 

1. In general I think this proposal is reasonable, but there does not seem to be consensus on it yet.
2. Is it important to allow this to be specified for individual dependencies? Major versions might be a big deal in some dependencies and not others.

elharo (changes requested):
This needs tests.

guyv:
@elharo regarding 2.: the exclude tag still works :-)

gmshake:
Cross link the issue https://issues.apache.org/jira/browse/MENFORCER-360

rfscholte:
I'm going to close this PR because there are conflicts and there are missing tests.

jira-importer:
Resolve #473

jira-importer:
Resolve #473

PR #79, opened 2020-09-13
[MENFORCER-361] optionally normalize line separators prior to checksum calculation

kwin:
Normalizing line separators is useful to properly support multi-platform development where text-based files are supposed to differ in their line separators (which should not have an impact on the checksum).

       pulling in other changes.
       commit message.


kwin:
Can someone have a look at this PR? Is there anything missing?

michael-o:
Can you describe the usecase for this?

kwin:
@michael-o Hopefully this is clear from https://issues.apache.org/jira/browse/MENFORCER-361. I want to use platform independent checksums for text files.

michael-o:
I see, do you already have a usecase for this?

kwin:
Yes, look at https://github.com/adobe/aem-project-archetype/pull/480 where this is currently leveraged with two different profiles: One for Windows and one for Linux/Mac OS.

gmshake:
Hi, @kwin 

The current enforcer `RequireFileChecksum` checks `file` checksum. For `text` files, line separator varies on different platform, OS X, Windows, Linux etc. For `binary` files, it is confusing to normalize the line separator. So, I'd propose we introduce a new `RequireTextFileChecksum` enforcer to make it clear.

kwin:
> The current enforcer RequireFileChecksum checks file checksum

Agree, but that includes both binary and text files. There is no hint in https://maven.apache.org/enforcer/enforcer-rules/requireFileChecksum.html that this should only be used for binaries.

> So, I'd propose we introduce a new RequireTextFileChecksum enforcer to make it clear.

I am fine with that as well, but honestly I don't see any benefit here. As both need configuration per file, one can just as well add another configuration flag (`normalizeLineSeparatorTo`). Why do you prefer two different rule names here?

gmshake:
@kwin

> Agree, but that includes both binary and text files. There is no hint in https://maven.apache.org/enforcer/enforcer-rules/requireFileChecksum.html that this should only be used for binaries.

We commonly calculate files's checksum regardless the content type, i.e.  we always treat them as `binary`. Since your usecase is special, normalizing line separator before calculating file's checksum, it's better not to confuse end user with the common usecase. 

Anyway, both works :)

michael-o:
I think this can be drastically simplified by saying:
* we have a `requireTextFileChecksum`
* Convert into a canonical text form (Unix style endings)

No further configuration will be required. This should be idiot-proof.

WDYT?

kwin:
Ok, will modify the PR accordingly. Notice though that the character encoding configuration is mandatory though (but I will use `${project.build.sourceEncoding}` by default)

michael-o:
Right, even on UTF-16 it will be four bytes for `\r\n`.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * @author Konrad Windszus
+ * @see RequireFileChecksum
+ */
+public class RequireTextFileChecksum
+    extends RequireFileChecksum
+{
+    private NormalizeLineSeparatorReader.LineSeparator normalizeLineSeparatorTo = LineSeparator.UNIX;
+
+    private Charset charsetEncoding;
```
michael-o, line 48: make it just `encoding` akin to the setter

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public void setNormalizeLineSeparatorTo( NormalizeLineSeparatorReader.LineSeparator normalizeLineSeparatorTo )
+    {
+        this.normalizeLineSeparatorTo = normalizeLineSeparatorTo;
+    }
+
+    public void setEncoding( String fileCharset )
```
michael-o, line 55: Arg too

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
```diff
@@ -0,0 +1,103 @@
...
+        // set defaults
+        if ( charsetEncoding == null )
+        {
+            // https://maven.apache.org/plugins/maven-resources-plugin/examples/encoding.html
+            try
+            {
+                String encoding = (String) helper.evaluate( "${project.build.sourceEncoding}" );
+                if ( StringUtils.isBlank( encoding ) )
+                {
+                    throw new EnforcerRuleException( "No encoding set for '${project.build.sourceEncoding}'" );
```
michael-o, line 73: This is wrong. It should be exactly like here: https://maven.apache.org/plugins/maven-resources-plugin/xref/org/apache/maven/plugins/resources/ResourcesMojo.html#L313

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
```diff
@@ -0,0 +1,103 @@
...
+            }
+        }
+        super.execute( helper );
+    }
+
+    @Override
+    protected String calculateChecksum()
+        throws EnforcerRuleException
+    {
+        try ( FileInputStream fileInputStream = new FileInputStream( this.file );
```
michael-o, line 90: Why so complex? What about `Files.newInputStream()` or `Files.newBufferedReader()`?

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/utils/NormalizeLineSeparatorReader.java
```diff
@@ -0,0 +1,177 @@
...
+{
+
+    private static final int EOL = -1;
+
+    /** 
+     * Type representing either Unix or Windows line separators
+     */
+    public enum LineSeparator
+    {
+        WINDOWS( "\r\n", null ), UNIX( "\n", '\r' );
```
michael-o, line 40: UNIX does not have `\r` as any line separator. Why is it there?

michael-o (changes requested):

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/utils/NormalizeLineSeparatorReader.java
michael-o: Ah thanks!

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
kwin: Good point, gonna simplify, left-over from refactoring.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
kwin: Fixed in https://github.com/apache/maven-enforcer/pull/79/commits/731853114214d25decf8c6a1eab726cd82783348.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
kwin: https://github.com/apache/maven-enforcer/pull/79/commits/731853114214d25decf8c6a1eab726cd82783348

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
kwin: https://github.com/apache/maven-enforcer/pull/79/commits/731853114214d25decf8c6a1eab726cd82783348

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireTextFileChecksum.java
kwin: https://github.com/apache/maven-enforcer/pull/79/commits/731853114214d25decf8c6a1eab726cd82783348

michael-o (changes requested):
Almost merge ready

michael-o:
Thanks, I will test tomorrow and merge. Ping me at 16:00 UTC if I forget this.

kwin:
The test failure in Windows, Java11 with GitHub Actions at https://github.com/apache/maven-enforcer/pull/79/checks?check_run_id=1260193340 doesn't seem related to my change (`org.apache.maven.plugins.enforcer.TestEvaluateBeanshell.testRuleCanExecuteMultipleThreads()`).
@michael-o Let me know if I should squash the commits.

michael-o:
I will have a look tomorrow at it.

kwin:
@michael-o Thanks for merging.

PR #82, opened 2021-01-12
update commons IO

elharo:
@michael-o last version that supports Java 7

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -143,7 +143,7 @@
       <dependency>
         <groupId>commons-io</groupId>
         <artifactId>commons-io</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -143,7 +143,7 @@
       <dependency>
         <groupId>commons-io</groupId>
         <artifactId>commons-io</artifactId>
-        <version>2.5</version><!-- same version as used by maven-shared-utils -->
```
michael-o, line 146: What does Shared Utils use now?

elharo: 2.6

PR #86, opened 2021-01-24
[MENFORCER-376] Add excludes/includes functionality to RequireJavaVendor rule

KroArtem:

       pulling in other changes.
       commit message.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
@@ -23,6 +23,8 @@ import org.apache.commons.lang3.SystemUtils;
 import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
 import org.apache.maven.enforcer.rule.api.EnforcerRuleHelper;
 
+import java.util.List;
+
 /**
  * This rule checks that the Java vendor is allowed.
  *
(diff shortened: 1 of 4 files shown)


enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
```diff
@@ -31,40 +33,112 @@
...
+    }
+
+    /**
+     * Gets the includes.
      *
-     * @param name the required name to set
+     * @return the includes
      */
-    public final void setName( String name )
+    public List<String> getIncludes()
```
rfscholte, line 112: this will break backwards compatibility.

Better to do something like

    public void setName(String name) 
    {
      this.includes = Collections.singletonList(name);
      getLog().warning("property name is deprecated, replace with includes.")
    }

This is just an example, please verify that the name still works.

KroArtem: What's up with backwards compatibility? This rule was not released in any version (M4 was not released), there is no information about this rule on site https://maven.apache.org/enforcer/enforcer-rules/index.html

From my point of view, this was specifically done to prevent breaking backward compatibility in future. Here is your comment as an example 😃 https://issues.apache.org/jira/browse/MENFORCER-338?focusedCommentId=17169274&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17169274

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
```diff
@@ -31,40 +33,112 @@
...
+    }
+
+    private void createException() throws EnforcerRuleException
+    {
+        String message = getMessage();
+        String error = "Vendor " + SystemUtils.JAVA_VENDOR + " is in a list of banned vendors: " + excludes;
+        StringBuilder sb = new StringBuilder();
+        if ( message != null )
+        {
+            sb.append( message ).append( System.lineSeparator() );
```
elharo, line 137: Just \n please. Exception messages aren't just for System.out.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
bmarwell: Probably just an oversight that this was never part of a release. Happens to all of us. :)

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
bmarwell: @KroArtem I think this might be the only fix needed. Rest looks fine to me.

KroArtem: ok, I'll fix it today. Just understood that documentation (`requireJavaVendor.apt.vm`) has to be updated too.

KroArtem:
I still have one question, what should happen if I define a list of `includes` but current vendor doesn't satisfy this list? With this implementation it doesn't fail. However, it seem to be aligned with `bannedPlugins` [rule](https://maven.apache.org/enforcer/enforcer-rules/bannedPlugins.html), citing it:

```
includes - a list of plugin artifacts to include. These are exceptions to the excludes. It is meant to allow wide exclusion rules with wildcards and fine tune using includes. If nothing has been excluded, then the includes have no effect. In otherwords, includes only subtract from artifacts that matched an exclude rule.
```

That's the only question left so far.

KroArtem:
@rfscholte , @bmarwell , @elharo , can I kindly ask you to review MR again?

elharo (changes requested):
The docs that support this feature should be added in this PR as well.

KroArtem:
> The docs that support this feature should be added in this PR as well.

@elharo , what exactly do you mean? I've already made corresponding changes in .apt.vm files. Is there something that should be updated too?

elharo:
Yes, those changes should be in this PR too.

KroArtem:
I may be silly but these changes ARE in the PR right now, still don't get what else should be done. Here is where changes are: https://github.com/apache/maven-enforcer/pull/86/files#diff-b2cfcc9c03ff6b43bade73782c36c14b900b63cb39f57bea78a40a6cf62ae259R33

enforcer-rules/src/site/apt/requireJavaVendor.apt.vm
```diff
@@ -53,7 +56,13 @@ Require Java Vendor
             <configuration>
               <rules>
                 <requireJavaVendor>
-                  <name>AdoptOpenJDK</name>
+                  <excludes>
+                    <exclude>Pivotal</exclude>
```
elharo, line 60: are these the actual vendor names we would see?

KroArtem: These are the names I took from jdk.dev

My current mvn -v returns the following:

```
Java version: 15.0.1, vendor: N/A, runtime: /usr/local/Cellar/openjdk/15.0.1/libexec/openjdk.jdk/Contents/Home
```

KroArtem:
@hboutemy , I see that you're also requested for a review, do you have any review comments?

KroArtem:
Based on a discussion in JIRA ticket, this PR should be reworked to support either excludes or includes, not both at the same time.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVendor.java
rfscholte: Confirmed

enforcer-rules/src/site/apt/requireJavaVendor.apt.vm
```diff
@@ -30,8 +30,11 @@ Require Java Vendor
    The following parameters are supported by this rule:
    
    * message - an optional message to the user if the rule fails.
-   
-   * name - The name of the allowed JDK vendor.
+
+   * excludes - a list of banned JDK vendors.
+
+   * includes - a list of allowed JDK vendors to include. These are exceptions to the excludes.
```
rfscholte, line 36: This is the opposite of how most Maven Plugins work: if includes is 1..5, these are the only allowed values. With excludes you can finetune the includes, e.g. excludes = 2,4. This would mean that the required value must be 1,3 or 5.
For consistency this should be changed.

rfscholte:
This PR was the base for https://github.com/apache/maven-enforcer/commit/3b47fa9627bdf96a9fb45aac31c655bc6cae5060

jira-importer:
Resolve #721

PR #97, opened 2021-05-27
MENFORCER-385 add rule to enforce that all transitive provided

kwin:
dependencies are contained in the runtime classpath

Still lacking ITs, but working properly already.
I would appreciate some feedback.


rfscholte:
https://issues.apache.org/jira/browse/MENFORCER-385 was closed as Won't Fix

jira-importer:
Resolve #500

PR #178, opened 2022-07-31
[MENFORCER-397] allow no rules

raupachz:
Hi there,

attached is another `up-for-grabs`. This PR allows no rules by adding a `failIfNoRules` parameter. Settings `failIfNoRules` to false will log a warning "No rules are configured" instead of throwing an exception. Default behaviour is to throw an exception if there are no rules configured. Like people are used to.

A suitable test case has been added.

Feedback is welcome.

       pulling in other changes.
       commit message.

### DIFF
--- a/maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
+++ b/maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
@@ -98,6 +98,14 @@ public class EnforceMojo
     private boolean failFast = false;
 
     /**
+     * Flag to fail the build if no rules are present
+     *
+     * @since 3.1.1
+     */
+    @Parameter( property = "enforcer.failIfNoRules", defaultValue = "true" )
+    private boolean failIfNoRules = true;
+
+    /**
      * Array of objects that implement the EnforcerRule interface to execute.
      */
     @Parameter( required = false )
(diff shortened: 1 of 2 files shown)


raupachz:
My bad. Let me check.

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
```diff
@@ -97,6 +97,12 @@
     @Parameter( property = "enforcer.failFast", defaultValue = "false" )
     private boolean failFast = false;
 
+    /**
+     * Set this to "true" to cause a failure if there are no rules to execute
+     */
+    @Parameter( property = "enforcer.failIfNoRules", defaultValue = "true" )
```
slawekjaranowski, line 103: Pleas add `@since`

Yes - I see that other parameters also has missing it ... we can fix for rest in separate PR

It is used for  documentation: https://maven.apache.org/enforcer/maven-enforcer-plugin/enforce-mojo.html

Maybe documentation - similar to other, eg:

```
Flag to allow executions without rules.
```

maven-enforcer-plugin/src/test/java/org/apache/maven/plugins/enforcer/TestEnforceMojo.java
```diff
@@ -296,6 +296,14 @@ public void testLoggingOnEnforcerRuleExceptionWithoutMessage()
         Mockito.verify( logSpy ).warn( Mockito.matches( ".* failed with message:" + System.lineSeparator() + "null" ) );
     }
 
+    @Test
+    public void testFailIfNoTests()
+            throws MojoExecutionException {
+        setupBasics( false );
+        mojo.setFailIfNoRules( false );
+        mojo.execute();
+    }
```
slawekjaranowski, line 305: please add some assertions, eg that `log.warn` was called only once with specific message and no more interactions on `log`

slawekjaranowski (changes requested):

slawekjaranowski (approved):

PR #180, opened 2022-08-11
[MENFORCER-422] Added externalRules rule

gastaldi:

       pulling in other changes.
       commit message.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/ExternalRules.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/ExternalRules.java
@@ -1,0 +1,152 @@
+package org.apache.maven.plugins.enforcer;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.maven.enforcer.rule.api.EnforcerRule;
+import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
+import org.apache.maven.enforcer.rule.api.EnforcerRuleHelper;
+import org.apache.maven.plugin.MojoExecution;
+import org.codehaus.plexus.classworlds.realm.ClassRealm;
+import org.codehaus.plexus.component.configurator.ComponentConfigurator;
+import org.codehaus.plexus.component.repository.exception.ComponentLookupException;
+import org.codehaus.plexus.configuration.PlexusConfiguration;
+import org.codehaus.plexus.configuration.xml.XmlPlexusConfiguration;
+import org.codehaus.plexus.util.xml.Xpp3DomBuilder;
+import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+
+/**
+ * An enforcer rule that will invoke rules from an external resource
+ *
+ * @author <a href="mailto:gastaldi@apache.org">George Gastaldi</a>
+ */
+public class ExternalRules extends AbstractNonCacheableEnforcerRule
+{
+    private static final String LOCATION_PREFIX_CLASSPATH = "classpath:";
+
+    /**
+     * The external rules location. If it starts with "classpath:", the resource is read from the classpath.
+     * Otherwise, it is handled as a filesystem path, either absolute, or relative to <code>${project.basedir}</code>
+     */
+    String location;
+
+    public ExternalRules()
+    {
+    }
+
+    public ExternalRules( String location )
+    {
+        this.location = location;
+    }
+
+    @Override
+    public void execute( EnforcerRuleHelper helper ) throws EnforcerRuleException
+    {
+        // Find descriptor
+        EnforcerDescriptor enforcerDescriptor = getEnforcerDescriptor( helper );
+        for ( EnforcerRule rule : enforcerDescriptor.getRules() )
+        {
+            rule.execute( helper );
+        }
+    }
+
+    /**
+     * Resolve the {@link EnforcerDescriptor} based on the provided {@link #descriptor} or {@link #descriptorRef}
+     *
+     * @param helper used to build the {@link EnforcerDescriptor}
+     * @return an {@link EnforcerDescriptor} for this rule
+     * @throws EnforcerRuleException if any failure happens while reading the descriptor
+     */
+    EnforcerDescriptor getEnforcerDescriptor( EnforcerRuleHelper helper )
+            throws EnforcerRuleException
+    {
+        try ( InputStream descriptorStream = resolveDescriptor( helper ) )
+        {
+            EnforcerDescriptor descriptor = new EnforcerDescriptor();
+            // To get configuration from the enforcer-plugin mojo do:
+            //helper.evaluate(helper.getComponent(MojoExecution.class).getConfiguration().getChild("fail").getValue())
+            // Configure EnforcerDescriptor from the XML
+            ComponentConfigurator configurator = helper.getComponent( ComponentConfigurator.class, "basic" );
+            configurator.configureComponent( descriptor, toPlexusConfiguration( descriptorStream ), helper,
+                    getClassRealm( helper ) );
+            return descriptor;
+        }
+        catch ( EnforcerRuleException e )
+        {
+            throw e;
+        }
+        catch ( Exception e )
+        {
+            throw new EnforcerRuleException( "Error while enforcing rules", e );
+        }
+    }
+
+    private InputStream resolveDescriptor( EnforcerRuleHelper helper )
+            throws ComponentLookupException, EnforcerRuleException
+    {
+        InputStream descriptorStream;
+        if ( location != null )
+        {
+            if ( location.startsWith( LOCATION_PREFIX_CLASSPATH ) )
+            {
+                String classpathLocation = location.substring( LOCATION_PREFIX_CLASSPATH.length() );
+                ClassLoader classRealm = getClassRealm( helper );
+                descriptorStream = classRealm.getResourceAsStream( classpathLocation );
+                if ( descriptorStream == null )
+                {
+                    throw new EnforcerRuleException( "Location '" + classpathLocation + "' not found in classpath" );
+                }
+            }
+            else
+            {
+                File descriptorFile = helper.alignToBaseDirectory( new File( location ) );
+                try
+                {
+                    descriptorStream = Files.newInputStream( descriptorFile.toPath() );
+                }
+                catch ( IOException e )
+                {
+                    throw new EnforcerRuleException( "Could not read descriptor in " + descriptorFile, e );
+                }
+            }
+        }
+        else
+        {
+            throw new EnforcerRuleException( "No location provided" );
+        }
+        return descriptorStream;
+    }
+
+    private static PlexusConfiguration toPlexusConfiguration( InputStream descriptorStream )
+            throws XmlPullParserException, IOException
+    {
+        return new XmlPlexusConfiguration( Xpp3DomBuilder.build( descriptorStream, "UTF-8" ) );
+    }
+
+    private ClassRealm getClassRealm( EnforcerRuleHelper helper ) throws ComponentLookupException
+    {
+        return helper.getComponent( MojoExecution.class ).getMojoDescriptor().getRealm();
+    }
+
+}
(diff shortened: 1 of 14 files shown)


maven-enforcer-plugin/src/it/projects/descriptors-local-always-pass/pom.xml
```diff
@@ -0,0 +1,56 @@
...
+        <version>@project.version@</version>
+        <executions>
+          <execution>
+            <id>test</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <Descriptors>
```
ppalaga, line 46: Shouldn't this be `<descriptors>` with lowercase `d`?

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * An enforcer rule that will invoke rules from an external resource
+ *
+ * @author <a href="mailto:gastaldi@apache.org">George Gastaldi</a>
+ */
+public class Descriptors extends AbstractNonCacheableEnforcerRule
+{
+    String descriptorRef;
+
+    String descriptor;
```
ppalaga, line 47: It looks I can only set one descriptor? It would be nice if I could set multiple descriptors from multiple sources and also be able to combine refs and local paths.

ppalaga (commented):
This would be great, thanks @gastaldi! Some comments inline. 

Another important feature for the future: be able to customize the descriptor before processing, ideally through an XSLT. Motivation: some artifacts that are banned in Quarkus might be required in Camel Quarkus. With XSLT I could remove those specific ones from Quarkus descriptor and keep the rest.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
```diff
@@ -0,0 +1,148 @@
...
+        }
+    }
+
+    private InputStream resolveDescriptor( EnforcerRuleHelper helper ) throws EnforcerRuleException
+    {
+        InputStream descriptorStream;
+        if ( descriptorRef != null )
+        {
+            descriptorStream = Thread.currentThread().getContextClassLoader()
+                    .getResourceAsStream( "enforcer-rules/" + descriptorRef + ".xml" );
```
ppalaga, line 98: I find this a bit lacking on self-explainability. I can imagine I'll need to google up every time what is the name of the magic folder. Can't we simply use a single xml element for both files and classpath resources, something like:

```
<descriptors>
    <descriptor>file:foo/bar.xml</descriptor><!-- relative to ${basedir} -->
    <descriptor>file:${maven.multiModuleProjectDirectory}/bar/baz.xml</descriptor><!-- absolute path within the source tree -->
    <descriptor>classpath:foo/bar.xml</descriptor><!-- class path resource -->
</descriptors>
```

maven-enforcer-plugin/src/it/projects/descriptors-local-always-pass/pom.xml
gastaldi: It doesn't matter for Plexus. The tests use capitalized names for rules so I used that

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
gastaldi: True, but you can also add another `<Descriptor>` rule, so that was an intended design choice 😉

gastaldi:
> Another important feature for the future: be able to customize the descriptor before processing, ideally through an XSLT. Motivation: some artifacts that are banned in Quarkus might be required in Camel Quarkus. With XSLT I could remove those specific ones from Quarkus descriptor and keep the rest.

@ppalaga because other transformation steps might be needed and because I wanted to keep it simple, I think this would be better served if the transformed file were already available as a separate dependency (eg. you could use the Maven dependency plugin to unpack the original rule XML, transform and include the transformed file in another dependency with another GAV), but we can discuss about that.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
gastaldi: That's an interesting approach. Just out of curiosity: Do we have other maven plugins doing the same? Just thinking how other plugins managed to solve the same problem

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
ppalaga: OK, thanks, that sounds good.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
ppalaga: I prefer doing it that way in my tools, but I do not know from top of my head whether any ASF plugins do it like that. 

If you want to stay similar with Assembly plugin, I can live with it.

Maybe we could try to find a better name for the top level element. How about `<externalRules>`?

ppalaga:
> you could use the Maven dependency plugin to unpack the original rule XML, transform and include the transformed file in another dependency with another GAV), but we can discuss about that.

Yes, I know, but it is a sh** lot of XML that I hoped to be able to avoid :)

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/Descriptors.java
gastaldi: +1, that sounds better, I suck at naming things 😀

gastaldi: > If you want to stay similar with Assembly plugin, I can live with it.

I don't mind changing it, I just want to avoid potential CVEs in the future 😀

gastaldi: @ppalaga now it's called `location` and it expects a file path or a resource from the classpath if it starts with `classpath:`
```xml
<externalRules>
    <location>classpath:enforcer-rules/quarkus.xml</location>
</externalRules>
```

You don't specify the `classpath:` prefix, it will respect the implementation in `org.apache.maven.plugin.PluginParameterExpressionEvaluator#alignToBaseDirectory`:

```xml
<externalRules>
    <location>classpath:enforcer-rules/quarkus.xml</location>
</externalRules>
```

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/ExternalRules.java
```diff
@@ -0,0 +1,151 @@
...
+ * An enforcer rule that will invoke rules from an external resource
+ *
+ * @author <a href="mailto:gastaldi@apache.org">George Gastaldi</a>
+ */
+public class ExternalRules extends AbstractNonCacheableEnforcerRule
+{
+    private static final String LOCATION_PREFIX_CLASSPATH = "classpath:";
+
+    /**
+     * The external rules location. If it starts with "classpath:", the resource is read from the classpath
```
ppalaga, line 49:
```suggestion
     * The external rules location. If it starts with "classpath:", the resource is read from the classpath.
     * Otherwise, it is handled as a filesystem path, either absolute, or relative to <code>${project.basedir}</code>
```
The "Otherwise" sentence is just my guess. Please correct it if necessary.

ppalaga (commented):
The code looks great! 
I made one JavaDoc suggestion inline. 
Otherwise, I wonder whether there is any Enforcer end user docs, where an `externalRules` example could be added?

gastaldi:
> Otherwise, I wonder whether there is any Enforcer end user docs, where an `externalRules` example could be added?

I added an example doc in the generated site

enforcer-rules/src/site/apt/externalRules.apt.vm
```diff
@@ -0,0 +1,72 @@
...
+  External Rules
+  ------
+  George Gastaldi
+  ------
+  August 2022
+  ------
+
+External Rules
+
+    This rule will evaluate rules from an external resource. It is useful for reusing enforcer rules from other dependencies.
```
ppalaga, line 28:
```suggestion
    This rule will evaluate rules from an external resource. It can be a classpath resource present in a <<<maven-enforcer-plugin>>> dependency or a local file.
```

enforcer-rules/src/site/apt/externalRules.apt.vm
```diff
@@ -0,0 +1,72 @@
...
+                        <id>enforce</id>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <ExternalRules>
+                                    <!-- if referencing a file, remove the 'classpath:' prefix -->
+                                    <location>classpath:enforcer/rules.xml</location>
+                                </ExternalRules>
```
ppalaga, line 61:
```suggestion
                                <ExternalRules>
                                    <!-- enforcer/rules.xml is supposed to be a classpath resource present -->
                                    <!-- in org.foo:foobar-rules dependency of maven-enforcer-plugin defined above -->
                                    <location>classpath:enforcer/rules.xml</location>
                                </ExternalRules>
                                <ExternalRules>
                                    <!-- You can add multiple <ExternalRules> elements if you need to enforce -->
                                    <!-- rules from multiple resources. -->
                                    <!-- src/build/rules.xml is a local file path relative to ${project.basedir} -->
                                    <!-- Protip: to set the same resource for all modules in your multimodule setup, --> 
                                    <!-- prefix the path with ${maven.multiModuleProjectDirectory}. That makes it -->
                                    <!-- relative to the root of your Maven module hierarchy. Make sure that there is -->
                                    <!-- a .mvn directory in your source tree root directory. -->
                                    <location>src/build/rules.xml</location>
                                </ExternalRules>
```

enforcer-rules/src/site/apt/externalRules.apt.vm
```diff
@@ -0,0 +1,72 @@
...
+                            <fail>true</fail>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+    [...]
+</project>
++---+
```
ppalaga, line 72: A `rules.xml` example would be nice here.

ppalaga (commented):
Thanks @gastaldi, some small docs suggestions inline.

ppalaga (approved):
Great work, thanks a lot, @gastaldi !

ppalaga:
Would you like to review, @michael-o ? Otherwise, I could merge.

enforcer-rules/src/site/apt/externalRules.apt.vm
```diff
@@ -0,0 +1,104 @@
...
+~~ KIND, either express or implied.  See the License for the
+~~ specific language governing permissions and limitations
+~~ under the License.
+
+  ------
+  External Rules
+  ------
+  George Gastaldi
+  ------
+  August 2022
```
michael-o, line 23: ISO format

enforcer-rules/src/site/apt/externalRules.apt.vm
```diff
@@ -0,0 +1,104 @@
...
+                                    <!-- enforcer/rules.xml is supposed to be a classpath resource present -->
+                                    <!-- in org.foo:foobar-rules dependency of maven-enforcer-plugin defined above -->
+                                    <location>classpath:enforcer/rules.xml</location>
+                                </ExternalRules>
+                                <ExternalRules>
+                                    <!-- You can add multiple <ExternalRules> elements if you need to enforce -->
+                                    <!-- rules from multiple resources. -->
+                                    <!-- src/build/rules.xml is a local file path relative to ${project.basedir} -->
+                                    <!-- Protip: to set the same resource for all modules in your multimodule setup, --> 
+                                    <!-- prefix the path with ${maven.multiModuleProjectDirectory}. That makes it -->
```
michael-o, line 68: Note that this property is an implementation detail and it should not be used in public

michael-o (changes requested):

enforcer-rules/src/site/apt/externalRules.apt.vm
gastaldi: So should we remove the pro-tip?

enforcer-rules/src/site/apt/externalRules.apt.vm
gastaldi: I copied the format from the `evaluateBeanshell.apt.vm` in this same directory, what ISO format are you expecting here?

michael-o: https://maven.apache.org/doxia/references/apt-format.html#title

enforcer-rules/src/site/apt/externalRules.apt.vm
michael-o: Either that or say that the behavior is undefined because this can break with every new release.

ppalaga: What a pity! Is there any official way to make a path relative to the root dir?

ppalaga:
Thanks, @gastaldi !

enforcer-rules/src/site/apt/externalRules.apt.vm
michael-o: There is an open JIRA issue

michael-o: @ppalaga  https://issues.apache.org/jira/browse/MNG-7038

ppalaga:
@michael-o do you happen to know whether an Enforcer release is planned already? Not sure where to look for that kind of info, although I am a committer.

michael-o:
> @michael-o do you happen to know whether an Enforcer release is planned already? Not sure where to look for that kind of info, although I am a committer.

We don't publish any plans since we are volunteers.

First of all, please complete the accompanying JIRA issue. Second, someone needs to tackle https://issues.apache.org/jira/browse/MENFORCER-407 then we can consider a release.

ppalaga:
Thanks for the info @michael-o !

jira-importer:
Resolve #720

jira-importer:
Resolve #720

PR #187, opened 2022-10-11
[MENFORCER-427] New rule to ban dynamic versions

kwin:

       pulling in other changes.
       commit message.

 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


michael-o:
Here are edge cases which I expect to work:
* [1.0.0] (hard 1.0.0)
* [1.0.0,1.0.0] (same as above), see https://issues.apache.org/jira/browse/MNG-7106

kwin:
- `[1.0.0,1.0.0]` is currently invalid and leads to `org.apache.maven.artifact.versioning.InvalidVersionSpecificationException: Range cannot have identical boundaries: [1.0,1.0]`.
- `[1.0]` leads to version resolving (although redundant) when used in a dependency and therefore always means overhead during build and dependency on metadata. I would rather ban this as well. This should be replaced by `1.0` to speed up the build!

kwin:
I opened https://issues.apache.org/jira/browse/MNG-7561 for same lower and upper bounds but it seems using a hard restriction always require resolving against metadata from local/remote repo, therefore at least optionally this pattern should be banned as well in this rule.

michael-o:
> * `[1.0.0,1.0.0]` is currently invalid and leads to `org.apache.maven.artifact.versioning.InvalidVersionSpecificationException: Range cannot have identical boundaries: [1.0,1.0]`.
> 
>     * `[1.0]` leads to version resolving (although redundant) when used in a dependency and therefore always means overhead during build and dependency on metadata. I would rather ban this as well. This should be replaced by soft requirement `1.0` to speed up the build! Is there already an issue related to this? According to https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification the fixed hard requirement may make sense for some edge cases, but as long as this leads to version resolving I would recommend not to rely on it.

@cstamas @gnodet What is your opinion on this.

michael-o:
> * `[1.0.0,1.0.0]` is currently invalid and leads to `org.apache.maven.artifact.versioning.InvalidVersionSpecificationException: Range cannot have identical boundaries: [1.0,1.0]`.

This is strange because 1.0 is resolved to this...I wonder why this range is logically invalid.

>     * `[1.0]` leads to version resolving (although redundant) when used in a dependency and therefore always means overhead during build and dependency on metadata. I would rather ban this as well. This should be replaced by soft requirement `1.0` to speed up the build! Is there already an issue related to this? According to https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification the fixed hard requirement may make sense for some edge cases, but as long as this leads to version resolving I would recommend not to rely on it.

Well, `[1.0]` exists for a reason. I think it is parsed to `[1.0,1.0]` then of course it incurs resolution. Hard for me to tell whether this should be blocked or not. If you want to block this, fine. Add a comment about the situation.

kwin:
> This is strange because 1.0 is resolved to this...I wonder why this range is logically invalid.

I proposed a fix for this attached to https://issues.apache.org/jira/browse/MNG-7106

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -0,0 +1,375 @@
...
+                    return !allowLatest;
+                }
+                else if ( versionConstraint.getVersion().toString().equals( RELEASE ) )
+                {
+                    return !allowRelease;
+                }
+                else if ( versionConstraint.getVersion().toString().endsWith( SNAPSHOT_SUFFIX ) )
+                {
+                    return !allowSnapshots;
+                }
```
michael-o, line 199: Isn't there a better way tthan relying on `#toString()`?

kwin:
>  I think it is parsed to [1.0,1.0] then of course it incurs resolution.

This is totally unexpected because `1.0` (soft requirement) is not resolved, see also my comment at https://github.com/apache/maven/pull/823#discussion_r993263931.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -0,0 +1,375 @@
...
+            if ( project == null )
+            {
+                throw new ExpressionEvaluationException( "${project} is null" );
+            }
+            RepositorySystemSession repoSession =
+                (RepositorySystemSession) helper.evaluate( "${repositorySystemSession}" );
+            if ( repoSession == null )
+            {
+                throw new ExpressionEvaluationException( "${repositorySystemSession} is null" );
+            }
```
michael-o, line 271: Please use `Objects` for these null checks.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -0,0 +1,375 @@
...
+            // get a new session to be able to tweak the dependency selector
+            newRepoSession = new DefaultRepositorySystemSession( repoSession );
+            remoteRepositories = (List<RemoteRepository>) helper.evaluate( "${project.remoteProjectRepositories}" );
+            repoSystem = helper.getComponent( RepositorySystem.class );
+        }
+        catch ( ExpressionEvaluationException eee )
+        {
+            throw new EnforcerRuleException( "Cannot resolve expression: " + eee.getCause(), eee );
```
michael-o, line 279: This duplicates the message. I think ` throw new EnforcerRuleException( "Cannot resolve expression", eee )` is enough.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -0,0 +1,375 @@
...
+        }
+        catch ( ComponentLookupException cle )
+        {
+            throw new EnforcerRuleException( "Unable to retrieve component RepositorySystem", cle );
```
michael-o, line 283: Only RS, not the rest?

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -0,0 +1,375 @@
...
+                ChoiceFormat dependenciesFormat = new ChoiceFormat( "1#dependency|1<dependencies" );
+                throw new EnforcerRuleException( "Found " + numViolations + " "
+                    + dependenciesFormat.format( numViolations )
+                    + " with dynamic versions. Look at the warnings emitted above for the details." );
+            }
+        }
+        catch ( DependencyCollectionException e )
+        {
+            throw new EnforcerRuleException( "Could not retrieve dependency metadata for project  : " + e.getMessage(),
+                                             e );
```
michael-o, line 312: Same here, don't duplicate.

enforcer-rules/src/site/apt/banDynamicVersions.apt.vm
```diff
@@ -0,0 +1,108 @@
...
+~~ KIND, either express or implied.  See the License for the
+~~ specific language governing permissions and limitations
+~~ under the License.    
+ 
+  ------
+  Ban Dynamic Versions
+  ------
+  Konrad Windszus
+  ------
+  October 2022
```
michael-o, line 23: Extended ISO date please.

enforcer-rules/src/site/apt/banDynamicVersions.apt.vm
```diff
@@ -0,0 +1,108 @@
...
+  
+  * <<allowLatest>> - if <<<true>>> dependencies with versions placeholder <<<LATEST>>> will not be banned.  Default is <<<false>>>.
+  
+  * <<allowRanges>> - if <<<true>>> versions ending with <<<-SNAPSHOT>>> will not be banned.  Default is <<<false>>>.
+  
+  * <<checkOptionals>> - if <<<true>>> also optional dependencies will be checked.  Default is <<<false>>>.
+  
+  * <<excludedScopes>> - the list of scopes to exclude;
+  
+  * <<ignores>> - a list of dependencies to ignore. The format is <<<groupId[:artifactId][:version][:type][:scope][:classifier]>>> where <<<artifactId>>>, <<<version>>>, <<<type>>>, <<<scope>>> and <<<classifier>>> are optional. Wildcards may be used to replace an entire or just parts of a section.
```
michael-o, line 52: This does not correspond to https://github.com/apache/maven-enforcer/pull/187/files#diff-9dd3c382c00a8c5fa752798d8f5e7bae9b72fe5d7454bebf08dc77d1244fe5b4R113.

michael-o (changes requested):

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
kwin: Is there? In Resolver API?

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -0,0 +1,375 @@
...
+
+    /**
+     * {@code true} if version ranges should be allowed
+     */
+    private boolean allowRanges;
+
+    /**
+     * {@code true} if also optional dependencies should be checked
+     */
+    private boolean checkOptionals;
```
gnodet, line 104: Why a default of `false` here ? In what sense are optional dependencies different wrt to version resolving ? If they aren't, I'd think the default should be also check those.
or maybe `excludeOptionals` if the above is true.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
michael-o: Honestly, I don't know. I am bit surprised. If there isn't one, this is bad. Just place a comment that there is no better solution then this is fine.

kwin: There is only https://github.com/apache/maven-resolver/blob/6e74e9ee279f827140553846b6421fe9b75d3ad9/maven-resolver-api/src/main/java/org/eclipse/aether/artifact/AbstractArtifact.java#L46 which is private. The only public API is when you deal with an artifact.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
kwin: The rest is injected via expressions!

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
michael-o: OK, accepted.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
michael-o: ah ok

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
kwin: Done in https://github.com/apache/maven-enforcer/pull/187/commits/182722b981b2257ec56bd8b34c3d21e813380cd5.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
kwin: Done in https://github.com/apache/maven-enforcer/pull/187/commits/182722b981b2257ec56bd8b34c3d21e813380cd5.

enforcer-rules/src/site/apt/banDynamicVersions.apt.vm
kwin: Done in https://github.com/apache/maven-enforcer/pull/187/commits/182722b981b2257ec56bd8b34c3d21e813380cd5.

enforcer-rules/src/site/apt/banDynamicVersions.apt.vm
kwin: This is actually the supported format, fixed the javadoc in https://github.com/apache/maven-enforcer/pull/187/commits/182722b981b2257ec56bd8b34c3d21e813380cd5.

gnodet (approved):

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -99,9 +100,15 @@
     private boolean allowRanges;
 
     /**
-     * {@code true} if also optional dependencies should be checked
+     * {@code true} if ranges having the same upper and lower bound like {@code [1.0]} should be allowed.
+     * Only applicable if {@link #allowRanges} is not set to {@code true}.
      */
-    private boolean checkOptionals;
+    private boolean allowRangesWithSameUpperAndLowerBound;
```
michael-o, line 106: That is huge. What about `allowRangesWithIdenticalBounds?

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
```diff
@@ -110,7 +117,8 @@
 
     /**
      * Specify the ignored dependencies. This can be a list of artifacts in the format
-     * <code>groupId[:artifactId][:version]</code>. Any of the sections can be a wildcard by using '*' (ie group:*:1.0)
+     * <code>groupId[:artifactId[:version[:type[:scope:[classifier]]]]]</code>. 
+     * Any of the sections can be a wildcard by using '*' (ie group:*:1.0).
```
michael-o, line 121: `.e.g {@code group:*:1.0}`

michael-o (changes requested):

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
kwin: only slighly shorter, but I changed to that in https://github.com/apache/maven-enforcer/pull/187/commits/2d82f485ff5c73a6052b189fa9488d02fa4b3cbb

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/BanDynamicVersions.java
kwin: fixed in https://github.com/apache/maven-enforcer/pull/187/commits/2d82f485ff5c73a6052b189fa9488d02fa4b3cbb.

michael-o (approved):

jira-importer:
Resolve #824

PR #192, opened 2022-11-22
remove specific IDE m2e files

olamy:
those files are specific to m2e they shouldn;t be stored in the Apache Maven sources!

 - [ ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MENFORCER) filed 
       pulling in other changes.
 - [ ] Each commit in the pull request should have a meaningful subject line and body.
 - [ ] Format the pull request title like `[MENFORCER-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [ ] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
 - [ ] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

 - [ ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


olamy:
cc @HannesWell @kwin

HannesWell (approved):
This should be safe as well, since this is also handled by M2E's default lifecycle-mappings.

https://github.com/eclipse-m2e/m2e-core/blob/master/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml#L51-L65

khmarbaise (approved):

slawekjaranowski (approved):

vtintillier:
Hello @olamy 

I think this is a regression for all users of Eclipse and maven-enforcer-plugin. We now have a warning in all our pom files because of this change:

image

It means this change forces all those projects to each find a way to ignore this warning. And for now I have not found a way to do it on my side.

The goal of this `META-INF/m2e/lifecycle-mapping-metadata.xml` file is just to declare whether it makes sense to run this plugin in Eclipse or not, so that not all projects have to declare it on their side.

Would you consider reverting this change to help all those projects? Especially given there was no maintenance on this file for 8 years in this project, so not really costing this project anything.

Thanks a lot

kwin:
@vtintillier Please raise an issue with m2e (at  https://github.com/eclipse-m2e/m2e-core/issues) as the action `ignore` in https://github.com/eclipse-m2e/m2e-core/blob/e3794059b1250729872daecf9c1ff0b35fc75daf/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml#L130 is the culprit.

vtintillier:
Hi @kwin 

looks like the ignore in the file you linked to has a different behavior than what we had with the `META-INF/m2e/lifecycle-mapping-metadata.xml` hosted in maven-enforcer-plugin itself. Before this file was removed, there would be no warning at all in Eclipse, and enforcer would not run as part of the Eclipse build.

Now enforcer still don't run, but we have this warning.

I asked about this in m2e too: https://github.com/eclipse-m2e/m2e-core/discussions/1229
But I think the easiest way forward is to add this file again.

Thanks

olamy:
Sorry for the disagreement but this file should have never been added.. 
But your problem should be solved by m2e plugin.
Imagine if every single IDE asks to host their configuration file.

PR #211, opened 2022-12-29
[MENFORCER-454] Deprecate 'rules' property and introduce 'enforcer.rules' as a replacement

psiroky:
This aligns the property names to us the `enforcer.` prefix.

Also updated the ITs to verify the rules are indeed getting fully replaced with the list from the property

       pulling in other changes.
       commit message.


maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
```diff
@@ -127,11 +127,22 @@ public class EnforceMojo extends AbstractMojo {
     private PlexusConfiguration rules;
 
     /**
-     * List of strings that matches the EnforcerRules to execute.
+     * List of strings that matches the EnforcerRules to execute. Use <code>enforcer.rules</code> property instead.
+     *
+     * @deprecated
```
slawekjaranowski, line 132: should be ` @deprecated descriptions ....`

please check `mvn site` output how documentation will be generated

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
```diff
@@ -127,11 +127,22 @@ public class EnforceMojo extends AbstractMojo {
...
 
+    /**
+     * List of strings that matches the EnforcerRules to execute. Replacement for the <code>rules</code> property.
+     * If both are specified, this one has a higher precedence.
+     *
+     * @since 3.2.0
+     */
+    @Parameter(required = false, property = "enforcer.rules")
+    private List<String> rulesToExecute;
+
```
slawekjaranowski, line 145: instead of adding next field we can try add `@Parameter` to setters method, so code can look like:

```java
@Parameter(required = false, property = "rules")
public void setCommandLineRules(String value) {
    this.rulesToExecute = value;
}

@Parameter(required = false, property = "enforcer .rules")
public void setRulesToExecute(String value) {
    this.rulesToExecute = value;
}
```

and we will have one field.

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
psiroky: Yeah, I had that previously, but the description is then not present in main table with the goal parameters, it only gets added to the "Parameter Details" list (which I missed before). I'll add here as well.

I did check the generated documentation. Looks good to me, after the suggested fix.

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
psiroky: I took a hybrid approach. I left the annotated field (so that we have the same setup as for the other fields) and added the setters to make sure the values are not getting overridden with empty lists (as the setter is getting called with empty list even if there is no configuration)

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
```diff
@@ -153,6 +155,25 @@ public class EnforceMojo extends AbstractMojo {
...
+        if (rulesToExecute != null && !rulesToExecute.isEmpty()) {
+            this.rulesToExecute = rulesToExecute;
+        }
+    }
+
+    public void setRulesToExecute(List<String> rulesToExecute) {
+        if (rulesToExecute != null && !rulesToExecute.isEmpty()) {
+            this.rulesToExecute = rulesToExecute;
+        }
+    }
```
psiroky, line 175: The only disadvantage of this approach is that we don't really have a direct control of which method is called first, so if both properties are specified the behavior could be potentially different (from what I was able to observe, the `setCommandLineRules` is getting called first, but I am not sure if that's guaranteed every time).

slawekjaranowski: Yes we don't have control which method will be called first.

but we can throw exception if `rulesToExecute` is already set with not empty list,
with information that only one parameter should be used - what do you think?

We need both setters form technical reason to have two parameters, but we can simply call `setRulesToExecute` from `setCommandLineRules` to have logic in one place - next my idea 😄

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
```diff
@@ -127,10 +127,12 @@ public class EnforceMojo extends AbstractMojo {
...
 
     /**
-     * List of strings that matches the EnforcerRules to execute.
+     * List of strings that matches the EnforcerRules to execute. Replacement for the <code>rules</code> property.
+     *
+     * @since 3.2.0
      */
-    @Parameter(required = false, property = "rules")
-    private List<String> commandLineRules;
+    @Parameter(required = false, property = "enforcer.rules")
```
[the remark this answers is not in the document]
psiroky: It seemed to me it would be more consistent to have it on the field, but I'll move it to the setter then.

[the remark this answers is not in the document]
slawekjaranowski: about consistent I thought that once rulesToExecute is set from setter with annotation and then from setter without annotation.

but of course it can be a my subjective opinion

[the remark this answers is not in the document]
psiroky: Yeah, I thought about is from the other fields (with `@Parameter`) point of view. And the other setter is just a glue to support the deprecated parameter. But it does not really matter that much. I moved the annotation to the setter.

psiroky:
@slawekjaranowski I applied your suggestions (hopefully correctly). I also added a `warn` log when using the old property. Please take another look when you get a chance.

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/EnforceMojo.java
```diff
@@ -153,6 +147,36 @@ public class EnforceMojo extends AbstractMojo {
...
+
+    /**
+     * List of strings that matches the EnforcerRules to execute.
+     *
+     * @deprecated Use <code>enforcer.rules</code> property instead
+     */
+    @Parameter(required = false, property = "rules")
+    @Deprecated
+    public void setCommandLineRules(List<String> rulesToExecute) throws MojoExecutionException {
+        getLog().warn("Detected the usage of property '-Drules' which is deprecated. Use '-Denforcer.rules' instead.");
```
slawekjaranowski, line 176: Maven 3.9 will take care about it and warn about deprecated items ... but can be now

jira-importer:
Resolve #841

PR #213, opened 2022-12-30
[MENFORCER-445] Include JAVA_HOME location in the Java rule failure messages

psiroky:

       pulling in other changes.
       commit message.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVersion.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVersion.java
@@ -76,6 +76,8 @@ public class RequireJavaVersion extends AbstractVersionEnforcer {
                 + " Build: " + detectedJdkVersion.getBuildNumber() + " Qualifier: "
                 + detectedJdkVersion.getQualifier());
 
+        setCustomMessageIfNoneConfigured(detectedJdkVersion, getVersion());
+
         enforceVersion(helper.getLog(), "JDK", getVersion(), detectedJdkVersion);
     }
 
@@ -111,5 +113,14 @@ public class RequireJavaVersion extends AbstractVersionEnforcer {
         String version = buffer.toString();
         version = StringUtils.stripEnd(version, "-");
         return StringUtils.stripEnd(version, ".");
+    }
+
+    private void setCustomMessageIfNoneConfigured(ArtifactVersion detectedJdkVersion, String allowedVersionRange) {
+        if (getMessage() == null) {
+            String message = String.format(
+                    "Detected JDK version %s (JAVA_HOME=%s) is not in the allowed range %s.",
+                    detectedJdkVersion, SystemUtils.JAVA_HOME, allowedVersionRange);
+            super.setMessage(message);
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


enforcer-rules/src/test/java/org/apache/maven/plugins/enforcer/TestRequireJavaVendor.java
```diff
@@ -56,25 +54,33 @@ public void matchingInclude() throws EnforcerRuleException {
...
         underTest.setIncludes(Collections.singletonList(NON_MATCHING_VENDOR));
         final EnforcerRuleHelper helper = EnforcerTestUtils.getHelper();
-        EnforcerRuleException e = assertThrows(EnforcerRuleException.class, () -> underTest.execute(helper));
-        assertThat(e.getMessage(), is(SystemUtils.JAVA_VENDOR + " is not an included Required Java Vendor"));
+
+        assertThatThrownBy(() -> underTest.execute(helper))
+                .isInstanceOf(EnforcerRuleException.class)
+                .hasMessage(
+                        "%s is not an included Required Java Vendor (JAVA_HOME=%s)",
+                        SystemUtils.JAVA_VENDOR, SystemUtils.JAVA_HOME);
```
psiroky, line 66: I replaced the exception assserts with AssertJ since I find those much easier to read and understand (and I noticed the project uses asserj elsewhere as well).

enforcer-rules/src/test/java/org/apache/maven/plugins/enforcer/TestRequireJavaVersion.java
```diff
@@ -89,30 +88,36 @@ void settingsTheJavaVersionAsNormalizedVersionShouldNotFail() throws EnforcerRul
...
     @Test
-    @Disabled
-    // TODO: Think about the intention of this test? What should it prove?
-    void thisShouldNotCrash() throws EnforcerRuleException {
-        RequireJavaVersion rule = new RequireJavaVersion();
-        rule.setVersion(SystemUtils.JAVA_VERSION);
-
-        EnforcerRuleHelper helper = EnforcerTestUtils.getHelper();
-        rule.execute(helper);
-        // intentionally no assertThat(...) because we don't expect and exception.
```
psiroky, line 115: This test has been disabled for years now and as far as I can tell it is not really testing anything that is not being already tested elsewhere, so I figured we should just delete it.

enforcer-rules/src/main/java/org/apache/maven/plugins/enforcer/RequireJavaVersion.java
```diff
@@ -76,6 +76,7 @@ public void execute(EnforcerRuleHelper helper) throws EnforcerRuleException {
                 + " Build: " + detectedJdkVersion.getBuildNumber() + " Qualifier: "
                 + detectedJdkVersion.getQualifier());
 
+        super.setMessage(prepareMessage(detectedJdkVersion, getVersion()));
```
slawekjaranowski, line 79: We override message which can be provided by rule configuration - is it intentionally?

psiroky: Hmm. So would it make sense to override the message only if it has been not provided by the rule configuration? Otherwise I am not sure how else would be get the additional info in.

slawekjaranowski: simply set if current message is null

PR #216, opened 2023-01-01
[MENFORCER-455] New Enforcer API

slawekjaranowski:
- New API based on JSR 330 components

---

       pulling in other changes.
       commit message.


slawekjaranowski:
TODO:
 - documentation for new API
 - move Built-In Rules to new API - can be done in separate PR

enforcer-api/src/site/apt/writing-a-custom-rule.apt.vm
```diff
@@ -29,57 +29,24 @@ Writing a custom rule
...
+%{snippet|id=project-pom|file=target/custom-rule-sample/pom.xml}
+
+  Note that the classloader is shared with the embedding <<<maven-enforcer-plugin>>> (a regular {{{https://maven.apache.org/guides/mini/guide-maven-classloading.html#Plugin_Classloaders}plugin classloader}})
+  and therefore the artifacts <<<org.apache.maven.enforcer:enforcer-api>>> are always loaded in the same version as the embedding <<<maven-enforcer-plugin>>>.
+
+  Custom rule artifacts should therefore only depend on <<<enforcer-api>>> and core Maven artifacts with <<<provided>>> scope (for details refer to {{{https://issues.apache.org/jira/browse/MNG-7097}MNG-7097}}).
+
+  The classes available from <<<enforcer-rules>>> must not be used from custom rules, as those are not considered API and may change in backwards-incompatible ways with every <<<maven-enforcer>>> version (even minor ones).
+
+  Another dependencies used by custom rule should be in compile scope.
```
kwin, line 46:
```suggestion
  Other dependencies used by custom rule at run time should have `compile` scope.
```

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
```diff
@@ -93,6 +112,16 @@ public List<EnforcerRuleDesc> createRules(PlexusConfiguration rules) throws Enfo
 
     private EnforcerRuleDesc createRuleDesc(String name, String implementation) throws EnforcerRuleManagerException {
 
+        // component name should always start at lowercase character
+        String ruleName = Character.toLowerCase(name.charAt(0)) + name.substring(1);
+
+        try {
+            return new EnforcerRuleDesc(ruleName, plexusContainer.lookup(EnforcerRuleBase.class, ruleName));
```
kwin, line 119: What is the default scope of Sisu components? How does it work, if the same rule is being used more than once with different parameters in parallel?
Shouldn't rules having the singleton scope lead to a warn at least to prevent race conditions with multi-threading?

What about allowing to reuse rule objects but always pass the current parametrisation as argument to the execute method?

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/AbstarctEnforcerLogger.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.maven.enforcer.rule.api.EnforcerLogger;
+import org.apache.maven.plugin.logging.Log;
+
+/**
+ * Base EnforcerLogger implementation
+ *
+ * @author Slawomir Jaranowski
+ * @since 3.2.0
+ */
+public abstract class AbstarctEnforcerLogger implements EnforcerLogger {
```
kwin, line 33:
```suggestion
public abstract class AbstractEnforcerLogger implements EnforcerLogger {
```

kwin (changes requested):

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
```diff
@@ -25,33 +25,52 @@
...
-    private Provider<MavenSession> sessionProvider;
+    private final Provider<MavenSession> sessionProvider;
 
-    @Inject
-    private Provider<MojoExecution> mojoExecutionProvider;
+    private final Provider<MojoExecution> mojoExecutionProvider;
+
+    private final ComponentConfigurator componentConfigurator;
+
+    private final PlexusContainer plexusContainer;
```
kwin, line 59: Hm, this is really legacy. New code should no longer rely on that but rather on https://www.eclipse.org/sisu/docs/api/org.eclipse.sisu.inject/reference/org/eclipse/sisu/inject/package-summary.html

kwin (changes requested):

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
kwin: We shouldn't support custom rules with legacy plexus annotations at all.

enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/AbstractEnforcerRule.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * This is the interface into the rule. This method should throw an exception
+     * containing a reason message if the rule fails the check. The plugin will
+     * then decide based on the fail flag and rule level if it should stop or just log the
+     * message as a warning.
+     *
+     * @throws EnforcerRuleException the enforcer rule exception
+     * @throws EnforcerRuleError     in order to brake a build immediately
+     */
+    public abstract void execute() throws EnforcerRuleException;
```
kwin, line 88: How to get hold of other Maven objects which are not Sisu components like 
- `RepositorySystemSession` or 
- current `MavenProject` ...?

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
slawekjaranowski: > What is the default scope of Sisu components? How does it work, if the same rule is being used more than once with different parameters in parallel? Shouldn't rules having the singleton scope lead to a warn at least to prevent race conditions with multi-threading?

Component annotated only by `@Named` has `per-lookup` strategy - every lookup return new instance of component.
Rule should not have `@Singleton` annotation.
We need new rule instance for each execution, each rule object is configured again - so we will override items in existing object.

> 
> What about allowing to reuse rule objects but always pass the current parametrisation as argument to the execute method (e.g. for custom rules having complex global initialisation)

Rule can depends on other component with `@Singleton` and manage it manually.

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
slawekjaranowski: So which Maven component should we use for lookup only JSR 330?

enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/AbstractEnforcerRule.java
slawekjaranowski: `RepositorySystemSession`, `MavenSesion`, `MavenProject` are scoped for current plugin relam, so can be injected in rule, as in example rule, and in `RequireNoRepositories`

enforcer-api/src/site/apt/writing-a-custom-rule.apt.vm
slawekjaranowski: done

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/AbstarctEnforcerLogger.java
slawekjaranowski: done

enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/AbstractEnforcerRule.java
kwin: To me it is not clear which objects can be injected. We need to document everything which is available in the Enforcer rule context.

slawekjaranowski: Every Maven component can be injected ... we can documented some of common but not all
Even more we dontt have a documentation for for every Maven components.

slawekjaranowski: We have the same situation for current API ... what component can be accessed by `EnforcerRuleHelper`

kwin: The point is: MavenProject or RepositorySystemSession are imho no plexus/Sisu components so I am wondering why they are available at all…

slawekjaranowski: Maven/Sisu magic - `PlexusContainer` is implemented by Sisu ...

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
slawekjaranowski: Even in new Api we use `PlexusContainer`

https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLookup.java

So now I don't see another way.

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleManager.java
slawekjaranowski: Even in new Api we use `PlexusContainer`

https://github.com/apache/maven/blob/master/maven-core/src/main/java/org/apache/maven/internal/impl/DefaultLookup.java

So now I don't see another way.

slawekjaranowski:
@kwin Do you accept my responses?
Can I do anything more with this PR?

Anybody else has any more remarks?
I would like to merge and go forward.

kwin (approved):
LGTM

jira-importer:
Resolve #822

jira-importer:
Resolve #822

PR #279, opened 2023-07-04
[MENFORCER-488] Add EnforcerLogger.isDebugEnabled()

kwin:


slachiewicz (approved):

slawekjaranowski (approved):
Maybe for completion we can add `is...Enabled` for other levels

slawekjaranowski (commented):
one missing `isWarnOrErrorEnabled`

kwin:
> one missing `isWarnOrErrorEnabled`

I would prefer to leave that out, as the semantics is not clear from the name. It could mean both

1. `isWarnEnabled() || isErrorEnabled()` or
2. `isWarnEnabled() && rule.getLevel() == EnforcerLevel.WARN || isErrorEnabled() && rule.getLevel() == EnforcerLevel.ERROR`

Although one could clarify that in the javadoc the chance for misinterpretation is quite high and I would not add it until we see a real use case.

slawekjaranowski (approved):

PR #282, opened 2023-08-22
[MENFORCER-490] Declare maven-enforcer-extension dependencies

elharo:
MENFORCER-490 is going to take multiple PRs to work through. This starts with a couple of the simpler fixes.

### DIFF
--- a/maven-enforcer-extension/pom.xml
+++ b/maven-enforcer-extension/pom.xml
@@ -35,6 +35,15 @@
       <artifactId>maven-core</artifactId>
     </dependency>
     <dependency>
+      <groupId>javax.inject</groupId>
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
As you see at build result - at least plexus-utils is needed in runtime ....

I suppose that can be a bug in exporting artifacts and classes by Maven classloader itself.

cstamas:
This is completely wrong PR: how is plexus-utils provided? It WAS provided in Maven 3.8 and before, but read release notes of Maven 3.9 (same for all 3.9.x version): https://maven.apache.org/docs/3.9.6/release-notes.html#potentially-breaking-core-changes-if-migrating-from-3-8-x

jira-importer:
Resolve #883

PR #289, opened 2023-09-09
[MENFORCER-490] Declare enforcer-rules dependencies

elharo:
This fixes the last easy dep declarations in this project. There are still a couple of issues that might point to bugs in the dependency analyzer. In particular, adding the dependency on javax.annotation:javax.annotation-api as done here results in an unused declared dependency warning, but taking it out results in a used undeclared dependency warning.

### DIFF
--- a/enforcer-api/pom.xml
+++ b/enforcer-api/pom.xml
@@ -24,7 +24,7 @@
   <parent>
     <groupId>org.apache.maven.enforcer</groupId>
     <artifactId>enforcer</artifactId>
-    <version>3.4.1-SNAPSHOT</version>
+    <version>3.4.2-SNAPSHOT</version>
   </parent>
 
   <artifactId>enforcer-api</artifactId>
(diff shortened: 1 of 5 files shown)


slawekjaranowski:
so what is the true about: ` javax.annotation:javax.annotation-api` is it used in project?

If is not used I would to not add only to have analyzer happy

elharo:
It is certainly used. I'm not sure why the dependency analyzer reports it as both used and unused. Likely a bug there I need to investigate:

```
$ grep -R "import javax.annotation" .
./enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/EnforcerRuleHelper.java:import javax.annotation.Nonnull;
./enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/EnforcerRule.java:import javax.annotation.Nonnull;
./enforcer-api/src/main/java/org/apache/maven/enforcer/rule/api/EnforcerRule.java:import javax.annotation.Nullable;
./maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleCache.java:import javax.annotation.PreDestroy;
```

jira-importer:
Resolve #883

jira-importer:
Resolve #883

PR #297, opened 2023-11-24
[MENFORCER-390] "requireFilesExist" no longer handles non-canonical paths

### LINKED ISSUE #501: [MENFORCER-390] "requireFilesExist" no longer handles non-canonical paths
**[Gene Smith](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=gene)** opened **[MENFORCER-390](https://issues.apache.org/jira/browse/MENFORCER-390?redirect=false)** and commented

With the commit to resolve MENFORCER-364, the rule "requireFileExists" checks that the canonical path of a file is the same as the absolute path.

But not all absolute paths are canonical.
* absolute paths can involve symbolic links
* and they are allowed to have parts which are relative
  * `/../`
  * `/./`

And when it fails to handle a path, it can report that a file does not exist, even though the local system will resolve the path.

A blunt solution might be three separate rules:
* requireFileExists
  * the provided path must resolve to a file (which may be a directory or link)
* requireCanonicalFileExists
  * the provided path must exist as a canonical file
* requireCasesenstiveFileExists
  * the provided path must file a file
  * the file name must have the same case (upper//lower) as the
  * the parts of the path from the file up must have the same case until they go through a symbolic link

I have used  the "nio" package to handle some of stuff before.  I will add a comment with some java code I would start with.  Since the outcome here is very dependent on the use case you pick, the java will be "meta code" with ??? where you have to know the use case to know the outcome.

but basically, with "nio" you can march up a path checking for symbolic links and such.

 

---

**Affects:** 3.0.0

**Remote Links:**
- [GitHub Pull Request #297
](https://github.com/apache/maven-enforcer/pull/297)

2 votes, 6 watchers

roadSurfer:
This reverts the change from MENFORCER-364 as this led to a regressions with symbolic links.

The fundamental issue is that there is no clean way to deal with case-sensitivity as OSs can have multuple filesystems mounted that follow different rules. Thus the simple `file.exists()` is, despite the limitations, probably best. Those requiring more stringent checks writing their own handling.

       pulling in other changes.
       commit message.


slawekjaranowski:
Ok, the simple `file.exists()` should be enough in normal usage ... 

I only think - if we don't break and special cases ... but agree more complicated check should be done optionally or by next rule

By the way we can mention in documentation a way how file existence is checked to be clear.

roadSurfer:
No, they are not intended. That will be my IDE doing its own thing, I'll
use a different editor and update.

roadSurfer:
That should be the whitepsace issues sorted. I also generated the site locally and it appeared OK to me.

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
```diff
@@ -26,7 +26,10 @@
 Require Files Don't Exist
 
   This rule checks that the specified list of files do not exist.
-
+  
+  The mounted filesystem(s) will dictate the case-sensitive rules when checking files. If you require checks that your
```
elharo, line 30: case-sensitivity

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
```diff
@@ -26,7 +26,10 @@
...
+  filesystem(s) do not support (e.g. a case-sensitive check on a case-insensitive filesystem) then you should consider
```
elharo, line 31: delete "then you should"

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
```diff
@@ -26,7 +26,10 @@
...
+  adding your own custom plugin that meets your specific needs.
```
elharo, line 32: your own --> a

enforcer-rules/src/site/apt/requireFilesExist.apt.vm
```diff
@@ -26,7 +26,10 @@
 Require Files Exist
 
   This rule checks that the specified list of files exist.
-
+  
+  The mounted filesystem(s) will dictate the case-sensitive rules when checking files. If you require checks that your
```
elharo, line 30: as above

enforcer-rules/src/site/apt/requireFilesSize.apt.vm
```diff
@@ -26,7 +26,10 @@
 Require File Size
 
   This rule checks that the specified list of files exist and are within the specified size range.
-
+  
+  The mounted filesystem(s) will dictate the case-sensitive rules when checking files. If you require checks that your
```
elharo, line 30: ditto

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -20,17 +20,16 @@
...
 
 import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
 
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.*;
```
elharo, line 32: project style does not use wildcard imports

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -107,15 +106,58 @@ void testEmptyFileListAllowNull() {
     @Test
     void testFileDoesNotExist() throws EnforcerRuleException, IOException {
```
elharo, line 107: I don't understand this test. It checks that we create a file, delete it, and then that the file doesn't exist? Consider renamingi

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -107,15 +106,58 @@ void testEmptyFileListAllowNull() {
...
+                .toFile();
+
+        try {
+            rule.setFilesList(Collections.singletonList(linkFile));
+            EnforcerRuleException e = assertThrows(EnforcerRuleException.class, rule::execute);
+            assertNotNull(e.getMessage());
+
+            linkFile.delete();
+            rule.execute();
+        } finally {
```
elharo, line 136: can we use try with resources for this?

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -107,15 +106,58 @@ void testEmptyFileListAllowNull() {
...
+        File canonicalFile = File.createTempFile("canonical_", null, temporaryFolder);
+        File linkFile = Files.createSymbolicLink(
+                        Paths.get(temporaryFolder.getAbsolutePath(), "symbolic.link"),
+                        Paths.get(canonicalFile.getAbsolutePath()))
+                .toFile();
+        canonicalFile.delete();
+        rule.setFilesList(Collections.singletonList(linkFile));
+
+        try {
+            rule.execute();
```
elharo, line 155: no exception is a pass?

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: This seems to just be a minor variation on the original test code, but I can change the name.
What do you think "testDeletedFileDetected"?
I'll update the symbolic link est to follow a similar formate.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: I don;'t think so, isn't that just for autoclosables?

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: Correct. This whole class it basically the inverse of `TestRequireFilesExist`.

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
roadSurfer: Updated

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
roadSurfer: Updated

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
roadSurfer: Updated

enforcer-rules/src/site/apt/requireFilesExist.apt.vm
roadSurfer: Updated

enforcer-rules/src/site/apt/requireFilesSize.apt.vm
roadSurfer: Updated

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: Updated, and in `TestRequireFilesExist` as well

enforcer-rules/src/site/apt/requireFilesExist.apt.vm
roadSurfer: Done. Apols.

enforcer-rules/src/site/apt/requireFilesSize.apt.vm
roadSurfer: Done. Apols.

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/files/RequireFilesExist.java
```diff
@@ -31,35 +30,11 @@ public final class RequireFilesExist extends AbstractRequireFiles {
     @Override
     boolean checkFile(File file) {
         // if we get here and the handle is null, treat it as a success
-        return file == null ? true : file.exists() && osIndependentNameMatch(file, true);
+        return file == null ? true : file.exists();
```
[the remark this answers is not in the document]
roadSurfer: Old habits die hard. :-)

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -107,12 +110,55 @@ void testEmptyFileListAllowNull() {
...
         File f = File.createTempFile("junit", null, temporaryFolder);
+        rule.setFilesList(Collections.singletonList(f));
+
+        // Check the file is detected as being present
+        EnforcerRuleException e = assertThrows(EnforcerRuleException.class, rule::execute);
+        assertNotNull(e.getMessage());
+
         f.delete();
 
         assertFalse(f.exists());
```
elharo, line 121: I think my confusion is that this is the only assert. It should be changed to an assumeFalse or removed.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
elharo: yes, renaming is a good idea

roadSurfer: Please give me a format of the new method name you'd prefer and I will ensure that all `test...Exists` and `test...DoesNotExist` methods are updated in all test classes.
Such a change will affect `TestRequireFilesSize`, `TestRequireFileChecksum` `TestRequireFilesDontExist`, and `TestRequireFilesExist`; this means touching classes that are unrelated to the bug fix in order to ensure naming consistency.

To be clear, I disagree with this but it is not my project and so I will follow your direction.

elharo: Don't change other methods not related to this PR. keep the PR focused.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: This is a quick negative check to ensure the file is detected and the test does not pass by fluke. It is consistent with checks in the other tests (throws an exception with a message).

And I am sorry, but I do no understand how `assumeFalse` is helpful here. If the file is not detected as being present and an exception thrown with a message, then that should be a hard failure as the rule is not working according to spec.

I can see a mix of checkin/not-checking for an unexpected exception, I can add and explicit check for one and add a `fail` if you would like? I don't think it would be appropriate to update every test though, just one directly required by this fix.

elharo: Test setup failure does not mean the test failed. It means the test wasn't run. That's the difference between assume and assert. JUnit reports these two cases differently.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
elharo: Better 9 wrong and 1 right than 10 wrong. That's the fundamental principle.

elharo: testSymbolicLinkDeletedDetected (spelling) but otherwise sounds good

roadSurfer: Done

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
```diff
@@ -26,7 +26,10 @@
 Require Files Don't Exist
 
   This rule checks that the specified list of files do not exist.
```
elharo, line 28: delete "list of"

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
```diff
@@ -26,7 +26,10 @@
...
-
+  
+  The mounted filesystem(s) will dictate the case-sensitivity rules when checking files. If checks are required that
```
elharo, line 30: delete "mounted" and consider getting rid of the parentheses or just changing this to "The filesystem" as is every file is on a single filesystem.

enforcer-rules/src/site/apt/requireFilesExist.apt.vm
```diff
@@ -27,22 +27,25 @@ Require Files Exist
 
   This rule checks that the specified list of files exist.
```
elharo, line 28: ditto

enforcer-rules/src/site/apt/requireFilesSize.apt.vm
```diff
@@ -27,26 +27,29 @@ Require File Size
 
   This rule checks that the specified list of files exist and are within the specified size range.
 
+  The mounted filesystem(s) will dictate the case-sensitivity rules when checking files. If you require checks that your
```
elharo, line 30: ditto

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -105,19 +109,62 @@ void testEmptyFileListAllowNull() {
...
 
     @Test
-    void testFileDoesNotExist() throws EnforcerRuleException, IOException {
+    void testDeletedFileDetected() throws EnforcerRuleException, IOException {
         File f = File.createTempFile("junit", null, temporaryFolder);
+        rule.setFilesList(Collections.singletonList(f));
+
+        // Check the file is detected as being present
+        EnforcerRuleException e = assertThrows(EnforcerRuleException.class, rule::execute);
+        assumeFalse(e.getMessage() == null);
```
elharo, line 118: This should be an assertNotNull

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
elharo: The wrong one was changed. Line 122 should be changed or better yet deleted. The point of the test is not whether f.delete() succeeds. You're not testing the delete method in java.io.File.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -105,19 +109,62 @@ void testEmptyFileListAllowNull() {
...
+        File canonicalFile = File.createTempFile("canonical_", null, temporaryFolder);
+        File linkFile = Files.createSymbolicLink(
+                        Paths.get(temporaryFolder.getAbsolutePath(), "symbolic.link"),
+                        Paths.get(canonicalFile.getAbsolutePath()))
+                .toFile();
+
+        rule.setFilesList(Collections.singletonList(linkFile));
+        // Check the link is detected as being present
+        EnforcerRuleException e = assertThrows(EnforcerRuleException.class, rule::execute);
+        assumeFalse(e.getMessage() == null);
```
elharo, line 139: assertNotNull

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
```diff
@@ -105,19 +109,62 @@ void testEmptyFileListAllowNull() {
...
+    void testSymbolicLinkTargetDeletedDetected() throws Exception {
+        File canonicalFile = File.createTempFile("canonical_", null, temporaryFolder);
+        File linkFile = Files.createSymbolicLink(
+                        Paths.get(temporaryFolder.getAbsolutePath(), "symbolic.link"),
+                        Paths.get(canonicalFile.getAbsolutePath()))
+                .toFile();
+
+        // Check the target is detected as being present
+        EnforcerRuleException e = assertThrows(EnforcerRuleException.class, rule::execute);
+        assumeFalse(e.getMessage() == null);
```
elharo, line 157: assertNotNull

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesExist.java
```diff
@@ -50,12 +56,34 @@ void testFileExists() throws Exception {
...
+                        Paths.get(temporaryFolder.getAbsolutePath(), "symbolic.link"),
+                        Paths.get(canonicalFile.getAbsolutePath()))
+                .toFile();
+        canonicalFile.delete();
+        rule.setFilesList(Collections.singletonList(linkFile));
+
+        try {
+            rule.execute();
+            fail("Should have received an exception");
+        } catch (Exception e) {
```
elharo, line 84: which exception should be thrown here? Be specific.

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
roadSurfer: Done

enforcer-rules/src/site/apt/requireFilesDontExist.apt.vm
roadSurfer: Done

enforcer-rules/src/site/apt/requireFilesExist.apt.vm
roadSurfer: Done

enforcer-rules/src/site/apt/requireFilesSize.apt.vm
roadSurfer: Done

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: Akready sorted based on earlier comment

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: Done

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesExist.java
roadSurfer: Done

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/files/TestRequireFilesDontExist.java
roadSurfer: Done

jemischi:
Please finish this patch-request. :)

slawekjaranowski:
> > Please finish this patch-request. :)
> 
> As far as I am aware, I made good all the required changes. I do not believe it is for me to mark things as "Resolved" but for the person who raised the issue or another team member.
> 
> @slawekjaranowski - Who are you awaiting feedback from? I do not believe it to be me.

@roadSurfer I missed totally this PR ... 
@elharo  had many comments .. more nits, many thread are open so I got lost in this 

I will look again and try to resolve it.

slawekjaranowski:
Thank you for this PR, and sorry it sat here for so long. It correctly identified a real regression, and the tests you wrote are what finally pinned it down.

The issue turned out to have two halves:

* **Non-canonical paths (`..`, `.`)** — fixed in #1011, which normalizes the path before comparing. The IT you added here passes on current `master` unchanged.
* **Symbolic links** — still broken, exactly as your `testSymbolicLinkExists` shows. `getCanonicalFile()` resolves a link to its target, so the link name never matched.

We went a slightly different route than the revert proposed here: #1031 keeps the case check but resolves with `toRealPath(NOFOLLOW_LINKS)` instead. That leaves the link alone while still correcting case on a case-insensitive filesystem, so symbolic links, `..` paths and the case check from #477 all work at the same time.

Your symbolic link tests for all three rules and the relative-path IT are included in #1031, and you are credited as co-author on the commit. One small note in case it is useful: in `testSymbolicLinkTargetDeletedDetected` the first `rule.execute()` ran before `setFilesList()`, so it was passing on the empty-file-list error rather than on the link check — moved the setup up in #1031.

Closing this in favour of #1031. Thanks again for the report and the tests.

PR #316, opened 2024-05-26
[MENFORCER-490] Remove unused dependency

elharo:
ITs passed locally so this doesn't appear to be needed. 

If it is needed we'll need to figure out how to properly exclude this from the dependency analyzer


slawekjaranowski:
There is an `extraArtifact` options for maven-invoker-plugin ... 
eventually dependency has an impact on module order

slawekjaranowski (approved):

PR #354, opened 2025-02-21
A grabbag of small simplifications found by IntelliJ

elharo:
There are also some likely bugs it pulled out, but I'll check those in another PR.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/BannedRepositories.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/BannedRepositories.java
@@ -96,18 +96,18 @@ public final class BannedRepositories extends AbstractStandardEnforcerRule {
     }
 
     // ----------------------------------------------------------------------
-    // Protected methods
+    // Package methods
     // ----------------------------------------------------------------------
 
-    protected void setBannedRepositories(List<String> bannedRepositories) {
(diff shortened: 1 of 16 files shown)


enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/BannedRepositories.java
```diff
@@ -96,18 +96,18 @@ public void execute() throws EnforcerRuleException {
     }
 
     // ----------------------------------------------------------------------
-    // Protected methods
+    // Package methods
```
gnodet, line 99: Changing visibility is an incompatibility.  Are we sure there are no custom rules overriding those methods ?

elharo: The class is final. That's why IntelliJ flagged these.

slawekjaranowski: yes for final class protected has no sense
maybe - private

slawekjaranowski (approved):

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/BannedRepositories.java
elharo: I tried private but these methods are used in the test classes.

PR #358, opened 2025-02-23
Remove unused collections

elharo:


Bukama:
Wondering if they should be put in the sets of the class variables?

Bukama (approved):

PR #367, opened 2025-05-30
Improve the performance of BannedDependencies

harrisric:
Improves performance in cases where there are lots of patterns and/or dependencies by preparing and re-using the predicates for testing each artifact.

- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/BannedDependencies.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/BannedDependencies.java
@@ -21,7 +21,10 @@ package org.apache.maven.enforcer.rules.dependency;
 import javax.inject.Inject;
 import javax.inject.Named;
 
+import java.util.function.Predicate;
+
 import org.apache.maven.artifact.Artifact;
+import org.apache.maven.enforcer.rule.api.EnforcerRuleException;
 import org.apache.maven.enforcer.rules.utils.ArtifactUtils;
 import org.apache.maven.execution.MavenSession;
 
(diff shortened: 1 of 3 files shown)


harrisric:
The same technique could be applied to `RequireReleaseDeps` and `BannedPlugins` if this is deemed a useful contribution.

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/utils/ArtifactMatcher.java
```diff
@@ -44,9 +44,10 @@ public final class ArtifactMatcher {
      * @author I don't know
      */
     public static class Pattern {
-        private String pattern;
+        private final String pattern;
 
-        private String[] parts;
+        private final String[] parts;
+        private final java.util.regex.Pattern[] partsRegex;
```
slawekjaranowski, line 50: why not use import here?

harrisric: The inner class this is part of is called `Pattern` so thus was to distinguish between them

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/utils/TestArtifactMatcher.java
```diff
@@ -125,6 +125,37 @@ void testMatch() {
...
+            assertEquals(true, matcher.match(artifactGood3));
+            assertEquals(false, matcher.match(artifactBad2));
+        }
+        long timeTakenNs = System.nanoTime() - startTime;
+        assertTrue(
+                timeTakenNs < 200_000_000,
+                String.format(
+                        "Time taken for repeated tests should be less than 200ms, but was: %dms",
+                        timeTakenNs / 1_000_000));
+    }
```
slawekjaranowski, line 157: I'm not sure about such performance test ....

can be flaky depends on system performance, like in last GH build we have:

```
Error:    TestArtifactMatcher.testMatcherPerformance:152 Time taken for repeated tests should be less than 200ms, but was: 200ms ==> expected: <true> but was: <false>

```

harrisric:
Build failure is on the new micro-benchmark added here:
https://github.com/apache/maven-enforcer/pull/367/files#diff-3dd05a1df96a3e0862d9e300a462a0d3c3efe8a1bd895d533a2746332ae02012R152

`TestArtifactMatcher.testMatcherPerformance:152 Time taken for repeated tests should be less than 200ms, but was: 200ms ==> expected: <true> but was: <false>`
Seems to have been exactly on the value that was allowed so would want to give it at least 10% margin for variation. The difference from the previous version was at least 50% faster.
Happy to take any suggestions on this as to whether this test is a good idea or an alternative approach.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/utils/TestArtifactMatcher.java
harrisric: @slawekjaranowski - I don't think that `ArtifactMatcher` is a very volatile area of the codebase so perhaps over-zealous to keep the performance test in. Therefore I'm happy to just take this as proof that that change is behaving as intended (i.e. you can run the test with/without the other changes and see the difference in performance) and remove it from the PR going forward (it's not doing anything exotic and so easy to recreate) to avoid any potential build flakiness.
@Bukama - thanks, that's a good recommendation I'll remember for the future.

Then a question remains whether to extend the technique to `RequireReleaseDeps `and `BannedPlugins` - happy to do that in this PR or a separate one, based on your preferences.

slawekjaranowski (approved):

PR #890, opened 2025-06-21
Improve example of custom rule

slawekjaranowski:
- use constructor for components injection
- use a sisu version from the project


cstamas (approved):

enforcer-api/src/custom-rule-sample/pom.xml
```diff
@@ -34,7 +34,6 @@
   <properties>
     <api.version>${project.version}</api.version>
     <mavenVersion>${mavenVersion}</mavenVersion>
-    <!-- use JDK 1.8 or 11 -->
     <maven.compiler.source>1.8</maven.compiler.source>
```
Bukama, line 37: shouldnt this be `release` always because the compiler plugin can handle to use source/target or release

PR #893, opened 2025-06-22
Introduce ruleName for rules configuration

slawekjaranowski:
We can define the same rule multiple times with different configuration
Each execution has the same default name, with ruleName we can differentiate them

### DIFF
--- a/maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleDesc.java
+++ b/maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleDesc.java
@@ -45,7 +45,8 @@ public final class EnforcerRuleDesc {
     }
 
     public String getName() {
-        return name;
+        String ruleName = rule.getRuleName();
+        return ruleName == null || ruleName.isEmpty() ? name : ruleName;
     }
 
     public EnforcerRuleBase getRule() {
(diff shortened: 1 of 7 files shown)


maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleDesc.java
```diff
@@ -45,7 +45,8 @@ public EnforcerRuleDesc(String name, EnforcerRuleBase rule) {
     }
 
     public String getName() {
-        return name;
+        String ruleName = rule.getRuleName();
+        return ruleName == null || ruleName.isEmpty() ? name : ruleName;
```
Bukama, line 49: Names with only whitespace are allowed as `isEmpty()` only checks `length == 0`. `isBlank()` (JDK 11?) implies `trim()` I suggest to disallow whitespace rule names.

Bukama (approved):
+0.5 (see comment, might be done in a separate PR, but should be considerd before first release of this features)

maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleDesc.java
slawekjaranowski: content is pom tags are trimed by default

PR #899, opened 2025-06-29
Remove unused javax.annotations dependency

elharo:


slawekjaranowski (approved):

pzygielo:
Maybe this could still be managed in top-level pom for
https://github.com/apache/maven-enforcer/blob/e2c841c8468834408b88d61ac893576352c058f7/maven-enforcer-plugin/src/main/java/org/apache/maven/plugins/enforcer/internal/EnforcerRuleCache.java#L21

and explicit version removed from
https://github.com/apache/maven-enforcer/blob/e2c841c8468834408b88d61ac893576352c058f7/maven-enforcer-plugin/pom.xml#L88
instead?

elharo:
PreDestroy is part of the JDK, no need to declare a separate dependency

pzygielo:
> PreDestroy is part of the JDK, 

Interesting. Since when?

PR #898, opened 2025-06-29
Rewrite CONTRIBUTING.md to use the Github issue tracker instead of JIRA

elharo:
We likely need to do this across all the repos that have migrated to github issues.

PR #911, opened 2025-07-11
Move contributing information into README

slawekjaranowski:
In the most Maven project we have such information in README, so to be consistent move it also here.

Use a similar format as in another project.

As continuation for: 
 - #898


Bukama (approved):

slawekjaranowski:
@pzygielo we can add a dedicated CONTRIBUTING file, but should be done in all project with similar content, the best with link to site.

PR #926, opened 2025-08-29
Update Version Ranges link in site.xml

### LINKED ISSUE #906: Documentation says that version ranges should not resolve to SNAPSHOT, but SNAPSHOT repos are still queried
### Affected version

3.6.0

### Bug description

The documentation for [maven-enforcer-plugin](https://maven.apache.org/enforcer/maven-enforcer-plugin/usage.html) has a link to [Version Ranges](https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges) that says:

> Resolution of dependency ranges should not resolve to a snapshot (development version) unless it is included as an explicit boundary.

However, in my observations, ALL configured repositories are queried for available versions, without regard to whether they are configured for snapshots or releases. It is possible that the quoted portion is still correct, and that SNAPSHOT versions are filtered out after retrieval. However, the expected behavior is that if these versions are to be excluded, then the repositories configured for snapshots only, should never be queried at all when resolving version ranges. This may be a bug elsewhere in Maven, but the linked documentation from the maven-enforcer-plugin should, at the very least, document this behavior, if it is expected.

Note: this behavior is particularly problematic, because repository.apache.org is configured as a snapshot repository by default in the Apache parent POM, and INFRA blocks when you access repository.apache.org with too many 404s. This means that too many version ranges anywhere in your dependency tree will get you banned, even if you never depended on any SNAPSHOT versions.

ctubbsii:
Remove link to outdated confluence page about version ranges, and use a link to a page on maven-enforcer's site instead.

This fixes #906 

- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/maven-enforcer-plugin/src/site/site.xml
+++ b/maven-enforcer-plugin/src/site/site.xml
@@ -39,7 +39,7 @@ under the License.
       <item name="Check specific rule via CLI" href="examples/check-specific-rule-via-cli.html"/>
     </menu>
     <menu name="Resources">
-      <item name="Version Ranges" href="https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges"/>
+      <item name="Version Ranges" href="../../enforcer/enforcer-rules/versionRanges.html"/>
     </menu>
   </body>
 </site>
--- a/maven-enforcer-plugin/src/site/site.xml
+++ b/maven-enforcer-plugin/src/site/site.xml
@@ -39,7 +39,7 @@ under the License.
       <item name="Check specific rule via CLI" href="examples/check-specific-rule-via-cli.html"/>
     </menu>
     <menu name="Resources">
-      <item name="Version Ranges" href="../../enforcer/enforcer-rules/versionRanges.html"/>
+      <item name="Enforcer Version Ranges" href="../../enforcer/enforcer-rules/versionRanges.html"/>
     </menu>
   </body>
 </site>
--- a/maven-enforcer-plugin/src/site/site.xml
+++ b/maven-enforcer-plugin/src/site/site.xml
@@ -39,7 +39,7 @@ under the License.
       <item name="Check specific rule via CLI" href="examples/check-specific-rule-via-cli.html"/>
     </menu>
     <menu name="Resources">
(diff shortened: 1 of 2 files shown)


ctubbsii:
The template's checkboxes for testing doesn't really apply here.

maven-enforcer-plugin/src/site/site.xml
```diff
@@ -39,7 +39,7 @@ under the License.
       <item name="Check specific rule via CLI" href="examples/check-specific-rule-via-cli.html"/>
     </menu>
     <menu name="Resources">
-      <item name="Version Ranges" href="https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges"/>
+      <item name="Version Ranges" href="../../enforcer/enforcer-rules/versionRanges.html"/>
```
Bukama, line 42: On other pages, e.g. https://maven.apache.org/enforcer/enforcer-rules/banDynamicVersions.html ,the "version ranges" links to: https://maven.apache.org/pom.html#Dependency_Version_Requirement_Specification

I suggest to bind all links to the same page.

maven-enforcer-plugin/src/site/site.xml
```diff
@@ -39,7 +39,7 @@ under the License.
       <item name="Check specific rule via CLI" href="examples/check-specific-rule-via-cli.html"/>
     </menu>
     <menu name="Resources">
-      <item name="Version Ranges" href="https://cwiki.apache.org/confluence/display/MAVENOLD/Dependency+Mediation+and+Conflict+Resolution#DependencyMediationandConflictResolution-DependencyVersionRanges"/>
+      <item name="Enforcer Version Ranges" href="../../enforcer/enforcer-rules/versionRanges.html"/>
```
slawekjaranowski, line 42: should be:

```suggestion
      <item name="Enforcer Version Ranges" href="../enforcer-rules/versionRanges.html"/>
```

to check again ... 😄

slawekjaranowski (approved):

maven-enforcer-plugin/src/site/site.xml
slawekjaranowski: @Bukama I think it is ok

Bukama (approved):

PR #930, opened 2025-09-12
Use SessionData for cache storage

slawekjaranowski:
Each Maven session should have a new cache.

Singleton can be shared between executions by Maven Daemon.


slawekjaranowski:
> Couldn't we just raise the minimum Maven requirement to 3.9 for enforcer?

It can  be the best and simple way ... but we discuss about 3.8.8 as minimum ....

So now I will use `SessionData` ...

PR #968, opened 2026-03-25
Add enforceBytecodeVersion rule based on mojohaus

cstamas:
This rule is very often used, and does not require anything special like new dependency, hence it should be among "built in" rules.

OTOH, it had issues, for example on larger multi projects did recheck same (potentially huge) JAR over and over again. It is now altered to not repeat same check (same options against same file) within same session.

This rule is 100% drop-in replacement for old uses (ITs are unmodified and just copied over), moreover rule itself has set priority that if enforce plugin gets updated with this rule, but the extra rules are present, the new rule will kick in. All the options are same as before, and rule name is unchanged as well.

Where it differs, and is reflected in ITs: is not so chatty as mojohaus one, it does not tell "skipped due scope", "skipped due regexp" etc, hence ITs are modified to simply assert for presence or absence of "[DEBUG] Analyzing GAV..." string instead.

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
@@ -202,6 +202,11 @@ public class EnforceBytecodeVersion extends AbstractStandardEnforcerRule {
      */
     private boolean strict = false;
 
+    /**
+     * Optimization to calculate same JAR with same options once per session, by default is enabled.
+     */
+    private boolean processOncePerSession = true;
+
     // ---
 
     /**
@@ -322,7 +327,7 @@ public class EnforceBytecodeVersion extends AbstractStandardEnforcerRule {
                 (ConcurrentMap<ChecksOptions, Boolean>) session.getRepositorySession()
                         .getData()
                         .computeIfAbsent(getClass().getSimpleName(), ConcurrentHashMap::new);
-        if (performedChecks.putIfAbsent(checksOptions, Boolean.TRUE) != null) {
+        if (processOncePerSession && performedChecks.putIfAbsent(checksOptions, Boolean.TRUE) != null) {
             // we already performed checks on this file with these parameters
             return null;
         }
--- a/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
@@ -18,25 +18,6 @@
  */
 package org.apache.maven.enforcer.rules.dependency;
 
-/*
- * Licensed to the Apache Software Foundation (ASF) under one
- * or more contributor license agreements.  See the NOTICE file
- * distributed with this work for additional information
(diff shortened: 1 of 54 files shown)


enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
```diff
@@ -0,0 +1,613 @@
...
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("1.6", 50);
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("6", 50);
+        // Java7
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("1.7", 51);
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("7", 51);
+        // Java8
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("8", 52);
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("1.8", 52);
+        // Java9
+        JDK_TO_MAJOR_VERSION_NUMBER_MAPPING.put("9", 53);
```
slawekjaranowski, line 95: Maps for jdk version > 52 is not requred, it will be computed
please look at methods `renderVersion` , `decodeMajorVersion`

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
```diff
@@ -0,0 +1,613 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
```
slawekjaranowski, line 2: License header is duplicated

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/EnforceBytecodeVersion.java
cstamas: version > 53, even ITs use notation like "1.9".

cstamas:
Applied PR comments and added new option `processOncePerSession` (by def true), that if configured to false this rule will behave as before, to be able to disable optimization if it comes in way for any reason.

slawekjaranowski:
We also need to add a documentation.

I don't know why one test fail on Jenkins ...

slawekjaranowski:
As it is a modified version not directly copied from mojohaus, maybe we can change tile of change to something like:

`Add enforceBytecodeVersion rule based on mojohaus`

To avoid questions about IP rights 😄

pom.xml
```diff
@@ -82,8 +82,11 @@
     <javaVersion>8</javaVersion>
     <mockito.version>4.11.0</mockito.version>
     <project.build.outputTimestamp>2025-09-28T21:22:14Z</project.build.outputTimestamp>
-    <!-- the same as in Maven  -->
+    <!-- all the same as in Maven  -->
     <resolver.version>1.9.25</resolver.version>
+    <version.plexus-utils>3.6.0</version.plexus-utils>
```
slawekjaranowski: 3.6.1 or 4.0.3 due to CVE
it is also in parent

slawekjaranowski (approved):

PR #979, opened 2026-05-14
Deps: Parent POM 48 and align deps

cstamas:
There were (most probably) dependabot that made
deps fall apart: Maven, Resolver and Sisu should
be upped in lockstep.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -160,12 +160,6 @@
         <version>${resolver.version}</version>
       </dependency>
       <dependency>
-        <groupId>org.eclipse.sisu</groupId>
-        <artifactId>org.eclipse.sisu.plexus</artifactId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -162,13 +162,13 @@
       <dependency>
         <groupId>org.eclipse.sisu</groupId>
         <artifactId>org.eclipse.sisu.plexus</artifactId>
-        <version>0.9.0.M2</version>
+        <version>1.0.0</version>
```
slawekjaranowski: it is in dep Managment in maven-parent

slawekjaranowski (approved):

pom.xml
cstamas: checked and dropped; tx

Issue #997, opened 2026-07-16
New rule: banSplitPackages — detect packages split across modules and artifacts


ascheman:
Follow-up to #995 (the `module-info` module rules: `requireExplicitModules`, `requireMinimalExports`, `banUnjustifiedOpens`).

## Problem

The Java module system requires every package to belong to exactly one module. When two modules on the module path
contain the same package, the boot layer refuses to resolve:

```
java.lang.module.ResolutionException: Module a.consumer contains package com.acme.p,
module a.provider exports package com.acme.p to a.consumer
```

Such a split package is easy to introduce by accident — most often when a build-time code generator, an unpacked or
shaded dependency, or a copied resource puts classes of a *foreign* package into the current project's output. The
failure surfaces late (at packaging, `jlink`, or first boot), and the message describes the symptom rather than the
build step that caused it.

Two further shapes make it worse:

- **Silent variant.** If the project does not `requires` the split module, the resolver never pulls it in — so it
  never complains. The build is green and the application boots, but a type or an endpoint that was supposed to come
  from that module is silently absent (a 404, a late `LayerInstantiationException`, a `NoClassDefFoundError`).
- **On the classpath.** A split package is legal there and the JVM never complains at all — but it is precisely what
  will break a later migration to the module path.

There is currently no Enforcer rule that catches any of these at build time.

## Proposed rule: `banSplitPackages`

Primary target is the **module path**, but the rule is useful on the **classpath** too (as a modularization aid), so
it covers both. Bind it after compilation (e.g. `process-classes`, or `verify`); it fails or warns when a package
produced by the current project also belongs to one of its dependencies.

Detection:

- **Project packages** — enumerate the packages actually present under the compiled output directory (the
  directories that hold `.class` files). This is intentionally based on the real output, not the compiled
  `module-info` descriptor: the offending classes are frequently added *after* `compile` (generation, copy), and the
  descriptor's `ModulePackages` attribute is only finalized at packaging time.
- **Dependency packages** — resolve the project's dependency artifacts with `java.lang.module.ModuleFinder`, whose
  `ModuleReference.descriptor().packages()` yields the full package set and transparently handles automatic modules;
  fall back to scanning jar entries for plain (non-modular) artifacts.
- **Intersect against the full declared-dependency set**, not only the modules the project `requires`. This is what
  lets the rule catch the **silent variant** that the resolver misses: the split provider is a real Maven dependency,
  so it is in the set even when the module graph would never pull it in.
[...]

PR #1000, opened 2026-07-19
Add banSplitPackages rule: module-path split-package detection

ascheman:
Implements the `banSplitPackages` rule proposed in #997 — the follow-up announced in #996.

> [!IMPORTANT]
> **Stacked on #996.** The first four commits are #996 (module-info reader + three rules); only the top three commits (`feat` / `test` / `docs` for `banSplitPackages`) are new here. Once #996 merges I will rebase onto master and the diff collapses to just the new rule. Draft until then.

## Rule

A package must belong to exactly one Java module — two modules on the module path containing the same package fail resolution (`java.lang.module.ResolutionException`), typically at packaging, `jlink`, or first boot, with a symptom-not-cause message. And if the offending module is not even `requires`d, the resolver never pulls it in: the build stays green while a type or endpoint is silently absent at runtime.

- **Project packages** come from **walking the output directory**, not the compiled descriptor — `ModulePackages` is only finalized at packaging time, and offending classes are frequently added *after* compile (code generation, copied resources). Both the classic layout and the **Maven 4 module source hierarchy** are supported (same `moduleOutputs()` discovery as #996).
- **Dependency packages** are read through `java.lang.module.ModuleFinder` (reflectively — same Java 8 baseline rationale as the reader in #996), which derives automatic-module names and reports complete package sets; artifacts the finder cannot model fall back to a plain class-file scan.
- The rule intersects against the **full declared-dependency set**, deliberately not only the `requires`d modules — this is what catches the silent variant.
- **Severity model:** overlap between two artifacts that both end up on the module path → **error**. An automatic module counts as "on the module path" only when a project module actually `requires` its derived name. Overlaps involving a classpath-only artifact are legal today and reported at `<classpathSeverity>` (default `warn`) as a modularization hazard.
- **Configuration:** `classpathSeverity` (`warn`/`error`/`ignore`), `allowedSplitPackages`, `ignoredModules`, `ignoredArtifacts`, `message`.
- **Non-goal:** a provider that is not a declared Maven dependency (pure runtime) is out of scope. The rule is a safety net, not a fix — the durable cure is placing generated/copied classes in the module that owns their package.

## Design note: dependency resolution

`ResolverUtil` is widened to `public`: the `enforce` mojo declares only `requiresDependencyCollection`, so artifact *files* are not resolved by default; the rule resolves them explicitly — the same approach `enforceBytecodeVersion` already uses.

## Tests

- 15 unit tests (Java 9+ gated; ASM fixtures in test scope only, extended with a JAR fixture writer for explicit/automatic/plain dependency artifacts).
- Four ITs under `maven-enforcer-plugin/src/it/projects/`: a classic two-artifact reactor failing on the silent variant, a clean pass, a classpath-overlap warn case (build succeeds, warning names the automatic module), and a Maven 4 module-source-hierarchy fail case (gated `4.0.0-rc-5+` / JDK 17+).
- Full `mvn clean verify -P run-its` green on Maven 3.10.0-rc-1 (158 ITs passed); the four new ITs additionally verified on Maven 4.0.0-rc-5.

---
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

(Replaces #999, which GitHub auto-closed when the head branch was renamed to `feature/997-ban-split-packages`.)

### DIFF
--- a/enforcer-rules/pom.xml
+++ b/enforcer-rules/pom.xml
@@ -159,6 +159,14 @@
       <artifactId>slf4j-simple</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- test only: generates module-info.class fixtures for the module rules; production code reads
+         module-info via java.lang.module.ModuleDescriptor (accessed reflectively, see JavaModuleInfoReader). -->
+    <dependency>
+      <groupId>org.ow2.asm</groupId>
+      <artifactId>asm</artifactId>
+      <version>9.7.1</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/ResolverUtil.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/ResolverUtil.java
@@ -54,7 +54,7 @@ import static org.apache.maven.artifact.Artifact.SCOPE_TEST;
  * Resolver helper class.
  */
 @Named
-class ResolverUtil {
+public class ResolverUtil {
 
     private final RepositorySystem repositorySystem;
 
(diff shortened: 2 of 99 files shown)


scordio:
Successfully tested locally on the AssertJ build, see output logs at https://github.com/assertj/assertj/pull/4331#issuecomment-5033659653.

PR #1002, opened 2026-07-26
fix: support groupId-only includes in RequireUpperBoundDeps rule (#998)

### LINKED ISSUE #998: RequireUpperBoundDeps does not support includes with only groupId
### Affected version

3.6.3

### Bug description

The docs for requireUpperBoundDeps state (emphasis mine):
> excludes - specify the dependencies that will be ignored. **The format is `groupId[:artifactId]`**
includes - specify the dependencies that will be checked. If not empty only these will be checked. **The format is `groupId[:artifactId]`**

... but in fact, the `artifactId` is *not* optional. We tried to narrow a plugin execution to only report on our team's first-party dependencies, by just including our own `groupId` in the execution, but when we tested it, it did not detect any of the known upper bound problems. We were only able to get it to work by specifying our `artifactId` values as well, which means we would have to specify every artifactId that we own. (This means, e.g., we can't reliably apply this enforcement in a central place without perpetually adjusting its includes every time we create a new shared component.)

I took a quick glance at RequireUpperBoundDeps.java and it appears to only check whether the include values *directly match* the visited dependencies, which is contrary to the docs (and prevents my use case).

The fix would be to make the artifactId truly optional when filtering dependencies for consideration, just like many of the other rules do: e.g., bannedDependencies works fine in this regard. Probably RequireUpperBoundDeps.java should be doing something like those rules: `Predicate<Artifact> shouldInclude = ArtifactUtils.prepareDependencyArtifactMatcher(includes)` instead of just a direct match comparison.

(My guess is that `excludes` probably has the same problem here, but I did not test it.)

waterWang:
### Summary

Fixes #998.

The `RequireUpperBoundDeps` rule documentation states that the `includes` parameter accepts `groupId[:artifactId]` format (with `artifactId` being optional), but the implementation only supported exact string matching via `List.contains()`. This meant specifying just a `groupId` (e.g., `com.example`) would not match dependencies like `com.example:my-lib`.

### Changes

- Changed `RequireUpperBoundDepsVisitor.includes` from `List<String>` to `Predicate<Artifact> includesMatcher`
- Used `ArtifactUtils.prepareDependencyArtifactMatcher()` to create the predicate — the same approach used by `BannedDependencies`, `RequireReleaseDeps`, and other rules that already support `groupId[:artifactId]` format
- Updated `visitEnter()` to use the predicate instead of direct string comparison

### Tests

- Added `testIncludesWithGroupIdOnly()` — verifies that `includes = ["default-group"]` correctly detects conflicts for dependencies with that groupId
- Added `testIncludesWithGroupIdOnlyExcludesOtherGroup()` — verifies that `includes = ["other-group"]` correctly ignores dependencies with a different groupId
- All 3 tests pass (1 existing + 2 new)

### DIFF
--- a/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/RequireUpperBoundDeps.java
+++ b/enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/RequireUpperBoundDeps.java
@@ -28,6 +28,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Predicate;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.artifact.versioning.ArtifactVersion;
@@ -203,7 +204,7 @@ public final class RequireUpperBoundDeps extends AbstractStandardEnforcerRule {
 
         private final ParentsVisitor parentsVisitor = new ParentsVisitor();
         private boolean uniqueVersions;
-        private List<String> includes = null;
+        private Predicate<Artifact> includesMatcher = null;
 
         public RequireUpperBoundDepsVisitor setUniqueVersions(boolean uniqueVersions) {
             this.uniqueVersions = uniqueVersions;
@@ -211,7 +212,9 @@ public final class RequireUpperBoundDeps extends AbstractStandardEnforcerRule {
         }
 
         public RequireUpperBoundDepsVisitor setIncludes(List<String> includes) {
-            this.includes = includes;
+            this.includesMatcher = includes != null && !includes.isEmpty()
+                    ? ArtifactUtils.prepareDependencyArtifactMatcher(includes)
+                    : null;
             return this;
         }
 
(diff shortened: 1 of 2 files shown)


waterWang:
Hi, friendly reminder! All CI checks passed on this PR. Could you take a look when you have a moment? Let me know if anything needs to be changed. Thanks!

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/dependency/RequireUpperBoundDeps.java
```diff
@@ -223,8 +226,11 @@ public boolean visitEnter(DependencyNode node) {
             DependencyNodeHopCountPair pair = new DependencyNodeHopCountPair(node, this);
             String key = pair.constructKey();
 
-            if (includes != null && !includes.isEmpty() && !includes.contains(key)) {
-                return true;
+            if (includesMatcher != null) {
```
elharo, line 229: what if instead of a null check we used a matcher that matches everything as the default?

elharo (changes requested):
run spotless

[22](https://github.com/apache/maven-enforcer/actions/runs/30221117065/job/96110076693?pr=1002#step:8:223)
Caused by: org.apache.maven.plugin.MojoExecutionException: The following files had format violations:
    src/test/java/org/apache/maven/enforcer/rules/dependency/RequireUpperBoundDepsTest.java
        @@ -18,6 +18,8 @@

slawekjaranowski:
please preserve line:

```
 I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004
```

form PR template in PR description

PR #1022, opened 2026-09-06
Improve RequireUpperBoundDeps error message

### LINKED ISSUE #937: RequireUpperBoundDeps error message is difficult to decipher
### Affected version

3.6.1

### Bug description

In certain situations the error reported by `RequireUpperBoundDeps` is hard to decipher.

For example, this error:
```
[ERROR] Rule 0: org.apache.maven.enforcer.rules.dependency.RequireUpperBoundDeps failed with message:
[ERROR] Failed while enforcing RequireUpperBoundDeps. The error(s) are [
[ERROR] Require upper bound dependencies error for org.springframework.data:spring-data-commons:3.4.5. Paths to dependency are:
[ERROR] +-com.example:enforcer-bug:0
[ERROR]   +-com.vaadin:vaadin-core:24.9.1
[ERROR]     +-com.vaadin:vaadin-core-internal:24.9.1
[ERROR]       +-com.vaadin:vaadin-core-components:24.9.1 (managed) <-- com.vaadin:vaadin-core-components:24.9.1
[ERROR]         +-com.vaadin:vaadin-combo-box-flow:24.9.1 (managed) <-- com.vaadin:vaadin-combo-box-flow:24.9.1
[ERROR]           +-org.springframework.data:spring-data-commons:3.4.5 (managed) <-- org.springframework.data:spring-data-commons:3.5.4
[ERROR] and
[ERROR] +-com.example:enforcer-bug:0
[ERROR]   +-com.vaadin:vaadin-core:24.9.1
[ERROR]     +-com.vaadin:vaadin-core-internal:24.9.1
[ERROR]       +-com.vaadin:vaadin-core-components:24.9.1 (managed) <-- com.vaadin:vaadin-core-components:24.9.1
[ERROR]         +-com.vaadin:vaadin-grid-flow:24.9.1 (managed) <-- com.vaadin:vaadin-grid-flow:24.9.1
[ERROR]           +-org.springframework.data:spring-data-commons:3.4.5 (managed) <-- org.springframework.data:spring-data-commons:3.5.4
[ERROR] ]
```

(Note, you have to look carefully - the version mismatch is between `3.4.5` and `3.5.4`).

If you inspect `com.vaadin:vaadin-combo-box-flow:24.9.1` and `com.vaadin:vaadin-grid-flow:24.9.1` you will see that they both declare this dependency:
```xml
<dependency>
    <groupId>org.springframework.data</groupId>
    <artifactId>spring-data-commons</artifactId>
    <version>3.5.4</version>
    <scope>compile</scope>
</dependency>
```

Yet the error message seems to be complaining that they are actually depending on `3.4.5` instead of `3.5.4` when you look at the `+-` child relationships.

The dependency on `3.4.5` is coming from the parent and you can see it in the `help:effective-pom`. But this is not at all obvious from the error message. In the error message, the `3.4.5` dependency is seemingly coming from "nowhere".

A contributor to this problem is that [the website for RequireUpperBoundDeps](https://maven.apache.org/enforcer/enforcer-rules/requireUpperBoundDeps.html) does not explain how to interpret the error messages.

Would it be possible to explain this kind of conflict more clearly? Thanks.

Below is a POM that reproduces this situation when you run `mvn validate`.

```xml
<project
  xmlns="http://maven.apache.org/POM/4.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <version>3.4.5</version>
    </parent>
    <groupId>com.example</groupId>
    <artifactId>enforcer-bug</artifactId>
    <name>Enforcer bug</name>
    <version>0</version>
    <packaging>war</packaging>

    <properties>
        <maven-enforcer-plugin.version>3.6.1</maven-enforcer-plugin.version>
        <vaadin.version>24.9.1</vaadin.version>
    </properties>

    <dependencies>
        <dependency>
            <groupId>com.vaadin</groupId>
            <artifactId>vaadin-core</artifactId>
        </dependency>
    </dependencies>

    <!-- Import vaadin-bom to define all the Vaadin dependency versions -->
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>com.vaadin</groupId>
                <artifactId>vaadin-bom</artifactId>
                <version>${vaadin.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <build>
        <plugins>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <version>${maven-enforcer-plugin.version}</version>
                <executions>
                    <execution>
                        <id>maven_enforcer</id>
                        <phase>validate</phase>
                        <goals>
                            <goal>enforce</goal>
                        </goals>
                        <configuration>
                            <rules>
                                <requireUpperBoundDeps>
                                    <excludes>
                                        <exclude>com.fasterxml.jackson.core:jackson-core</exclude>
                                        <exclude>com.fasterxml.jackson.core:jackson-databind</exclude>
                                        <exclude>com.fasterxml.jackson.datatype:jackson-datatype-jsr310</exclude>
                                        <exclude>com.google.guava:guava</exclude>
                                        <exclude>commons-codec:commons-codec</exclude>
                                        <exclude>io.netty:netty-all</exclude>
                                        <exclude>io.projectreactor.netty:reactor-netty</exclude>
                                        <exclude>org.apache.commons:commons-lang3</exclude>
                                        <exclude>org.apache.commons:commons-text</exclude>
                                        <exclude>org.jspecify:jspecify</exclude>
                                        <exclude>org.slf4j:slf4j-api</exclude>
                                    </excludes>
                                </requireUpperBoundDeps>
                            </rules>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
</project>
```

dhruv9b:
## Description

Improve the `RequireUpperBoundDeps` error message when a dependency version is managed to a different version.

Previously, managed dependencies were displayed like:

`org.example:childA:1.0.0 (managed) <-- org.example:childA:2.0.0`

The meaning of the arrow notation is not immediately clear, making it difficult to determine which version was selected and which version was requested.

This changes the message to:

`org.example:childA:1.0.0 (managed; requested org.example:childA:2.0.0)`

This makes the relationship between the managed version and the requested version explicit.

### Changes

* Updated `RequireUpperBoundDeps` to use clearer managed-version error message wording.
* Added a regression test covering the managed-version case.
* Extended `DependencyNodeBuilder` to create dependency nodes with a premanaged version for testing.
* Added documentation explaining the `managed; requested` notation.

### Testing

* `mvn verify` — passed
* `mvn -Prun-its verify` — passed

Fixes #937

---

Following this checklist to help us incorporate your contribution quickly and easily:

* [x] Your pull request should address just one issue, without pulling in other changes.

* [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.

* [x] Each commit in the pull request should have a meaningful subject line and body.

* [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.

* [x] Run `mvn verify` to make sure basic checks pass.

* [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

* [x] I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004.

* [ ] In any other case, please file an Apache Individual Contributor License Agreement.


slawekjaranowski (changes requested):
Please remove commit `[ENFORCER-939] Support enforcer.skip without a POM` from PR

dhruv9b:
> Please remove commit `[ENFORCER-939] Support enforcer.skip without a POM` from PR

Done

slawekjaranowski (changes requested):
Thanks for working on this! A few remarks before we can merge.

### 1. The notation should be consistent with `dependency:tree`

Maven already has an established notation for a managed version — `maven-dependency-tree` (`VerboseDependencyNode`) renders it in the verbose dependency tree as:

```
+- org.example:childA:jar:1.0.0:compile (version managed from 2.0.0)
```

Rather than introducing a third variant, let's reuse exactly that, so a user sees the same wording in `mvn dependency:tree -Dverbose` and in the enforcer output:

```
+-org.example:childA:1.0.0 (version managed from 2.0.0)
```

This affects the rule, the unit test and `require-upper-bound-dependencies-managed_failure/verify.groovy`.

### 2. Please add a new test instead of modifying the existing one

`RequireUpperBoundDepsTest#testRule` currently covers the plain conflict between two dependency paths and asserts that both versions show up in the message. After adding `withPremanagedVersion("2.0.0")` to the first child, that node alone already triggers the conflict and the assertion no longer verifies the two-path output — so the original case is effectively no longer covered.

Could you please restore `testRule` to its previous form and add a separate test (e.g. `testManagedVersion`) for the managed case? Then both behaviours stay covered. `DependencyNodeBuilder#withPremanagedVersion` is a good addition and can stay as is.

### 3. Documentation text is placed inside the code block

In `enforcer-rules/src/site/markdown/requireUpperBoundDeps.md.vm` the new sentence was inserted before the closing fence, so it will be rendered as part of the sample log output. It has to go after the closing ```` ``` ````.

Also, the sample log in that block contains no managed dependency at all, so the explanation has nothing to refer to. A short, separate example showing a managed dependency would be much clearer.

### 4. Unrelated changes

`buildErrorMessage()` only gained blank lines — please drop them to keep the diff focused on one change.

dhruv9b:
> Thanks for working on this! A few remarks before we can merge.
> 
> ### 1. The notation should be consistent with `dependency:tree`
> Maven already has an established notation for a managed version — `maven-dependency-tree` (`VerboseDependencyNode`) renders it in the verbose dependency tree as:
> 
> ```
> +- org.example:childA:jar:1.0.0:compile (version managed from 2.0.0)
> ```
> 
> Rather than introducing a third variant, let's reuse exactly that, so a user sees the same wording in `mvn dependency:tree -Dverbose` and in the enforcer output:
> 
> ```
> +-org.example:childA:1.0.0 (version managed from 2.0.0)
> ```
> 
> This affects the rule, the unit test and `require-upper-bound-dependencies-managed_failure/verify.groovy`.
> 
> ### 2. Please add a new test instead of modifying the existing one
> `RequireUpperBoundDepsTest#testRule` currently covers the plain conflict between two dependency paths and asserts that both versions show up in the message. After adding `withPremanagedVersion("2.0.0")` to the first child, that node alone already triggers the conflict and the assertion no longer verifies the two-path output — so the original case is effectively no longer covered.
> 
> Could you please restore `testRule` to its previous form and add a separate test (e.g. `testManagedVersion`) for the managed case? Then both behaviours stay covered. `DependencyNodeBuilder#withPremanagedVersion` is a good addition and can stay as is.
> 
> ### 3. Documentation text is placed inside the code block
> In `enforcer-rules/src/site/markdown/requireUpperBoundDeps.md.vm` the new sentence was inserted before the closing fence, so it will be rendered as part of the sample log output. It has to go after the closing `` ``` ``.
> 
> Also, the sample log in that block contains no managed dependency at all, so the explanation has nothing to refer to. A short, separate example showing a managed dependency would be much clearer.
> 
> ### 4. Unrelated changes
> `buildErrorMessage()` only gained blank lines — please drop them to keep the diff focused on one change.

My bad I was busy last week,I have changed the things as you asked.

PR #892, opened 2025-06-22
`RequireSameVersions` rule: consider entries in `<plugins>`  when analyzing build- and report-plugins

turing85:
Resolves #372 

### `RequireSameVersion` rule
 if plugins are analyzed, entries under `<plugins>` must be considered f
Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
- [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #1024, opened 2026-09-12
Fix/916 require same versions for managed report plugins

### LINKED ISSUE #916: RequireSameVersion failed even with defined versions of failsafe,surefire,surefire-report-plugin
### Affected version

3.6.1,3.6.0

### Bug description

I have defined the maven-enforcer-rule `RequireSameVersions` to keep the versions of 
* maven-surefire-plugin
* maven-failsafe-plugin
* maven-surefire-report-plugin
in sync. in my `parent-pom.xml` in the example project.

You can reproduce the failure with this [example project](https://github.com/khmarbaise/maven-enforcer-plugin-requireSameVersion) by running `mvn verify`:
```
[INFO] --- enforcer:3.6.1:enforce (enforce-maven) @ enforcer-test ---
[INFO] Rule 1: org.apache.maven.enforcer.rules.version.RequireMavenVersion passed
[INFO] Rule 2: org.apache.maven.enforcer.rules.RequirePluginVersions passed
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.255 s
[INFO] Finished at: 2025-07-31T20:52:53+02:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.6.1:enforce (enforce-maven) on project enforcer-test: 
[ERROR] Rule 0: org.apache.maven.enforcer.rules.RequireSameVersions failed with message:
[ERROR] Found entries with different versions
[ERROR] Entries with version 3.5.3
[ERROR] - org.apache.maven.plugins:maven-surefire-plugin:maven-plugin (buildPlugin)
[ERROR] Entries with version RELEASE
[ERROR] - org.apache.maven.plugins:maven-surefire-report-plugin:maven-plugin (reportPlugin)
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

If you change the plugin version for maven-enforcer-plugin in `parent-pom.xml` to 3.6.0 the failure will look the same.

If you change maven-enforcer-plugin version to `3.5.0` there is no failure anymore. The first occurence is obviously in `3.6.0`.

If you omit the `reporting` area from `parent-pom.xml` you will no issue at all.

The same behaviour can be observed with Maven versions like this:
* 4.0.0-rc4
* 3.9.11
* 3.9.10
* 3.9.9
* 3.8.8

It is not related to maven core it is clearly related to the maven-enforcer-plugin.

dhruv9b:
## Description

Fixes #916.

`RequireSameVersions` can incorrectly report different versions for a report plugin when the plugin version is managed through `<pluginManagement>`.

When a report plugin is declared without an explicit version, Maven can expose its artifact with version `RELEASE`, even though the plugin has a version defined in the project's `<pluginManagement>`.

For example, this can result in `RequireSameVersions` comparing:

```text
maven-surefire-plugin        3.5.3
maven-surefire-report-plugin RELEASE
```

and incorrectly reporting a version mismatch.

### Changes

The report-plugin version lookup now:

* Uses an explicitly configured version from the reporting plugin when available.
* Falls back to the version configured in the project's build `<pluginManagement>`.
* Falls back to the artifact version/base version when no configured or managed version is available.

The existing handling of `<plugins>`, `<buildPlugins>`, and `<reportPlugins>` is preserved.

### Tests

Added a unit regression test covering a report plugin whose artifact version is `RELEASE` while its managed version is `3.5.3`.

Added an integration test reproducing the issue with a report plugin whose version is provided through `<pluginManagement>`.

The existing `RequireSameVersions` tests and integration tests were also run to ensure the changes do not regress the behavior introduced by #892, where entries in `<plugins>` are considered when checking build and report plugins.

Fixes #916
- [x] I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004


pzygielo:
- I've re-seen similar in https://github.com/s4u/pgpverify-maven-plugin/issues/720.

Then I've re-found https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Configuring_Reporting_Plugins

> > **Important Note**:
> > ...
> > For reporting plugins, specify each version in the `<reporting><plugins/></reporting>` element (and in the `<build><pluginManagement/></build>` element too).

That has re-surprised me a lot, to encourage such duplication...

So for:

> Added an integration test reproducing the issue with a report plugin whose version is provided through `<pluginManagement>`.

I'd like to confirm - report plugin version IS NOT provided by `<pluginManagement>`, is it?

dhruv9b:
> Contributor

Yes, the report plugin version is provided through `<build><pluginManagement>`, but it is intentionally not specified explicitly under `<reporting><plugins>`.

This matches the configuration from the original #916 reproducer. Maven allows the reporting plugin version to be resolved from `<build><pluginManagement>` when it is not specified in `<reporting>`.

The issue is that `RequireSameVersions` was seeing the report artifact with version `RELEASE`, while the corresponding build plugin was `3.5.3`, even though the report plugin was managed at `3.5.3`.

So the integration test intentionally leaves the reporting version unspecified to reproduce that exact case. The fix uses the managed version when no explicit reporting version is present.

pzygielo:
I'm confused for following reasons:
- as also noted in https://github.com/apache/maven-enforcer/issues/916#issuecomment-3146433534 - `project.getReportArtifacts()` - returns plugins w/o version if there is none specified in `<reporting>` section (exactly this causes s4u/pgpverify-maven-plugin#720 to use `RELEASE`/the latest, which is not the same as the one used in `<pluginManagement>`)
- [Configuring_Reporting_Plugins](https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Configuring_Reporting_Plugins) says to provide explicit version in `<reporting>`
- the actual behavior of Maven itself is indeed to:
> Maven allows the reporting plugin version to be resolved from <build><pluginManagement> when it is not specified in <reporting>.

Can you link to some documentation that support:
> Maven allows the reporting plugin version to be resolved from <build><pluginManagement> when it is not specified in <reporting>

?

pzygielo:
(I'm not challenging this PR in any way, just using this, perhaps not quite proper channel, to learn more about that case.)

dhruv9b:
> I'm confused for following reasons:
> 
> * as also noted in [RequireSameVersion failed even with defined versions of failsafe,surefire,surefire-report-plugin #916 (comment)](https://github.com/apache/maven-enforcer/issues/916#issuecomment-3146433534) - `project.getReportArtifacts()` - returns plugins w/o version if there is none specified in `<reporting>` section (exactly this causes [Incorrect (overzealous) version of [managed] reporting plugin being verified s4u/pgpverify-maven-plugin#720](https://github.com/s4u/pgpverify-maven-plugin/issues/720) to use `RELEASE`/the latest, which is not the same as the one used in `<pluginManagement>`)
> * [Configuring_Reporting_Plugins](https://maven.apache.org/guides/mini/guide-configuring-plugins.html#Configuring_Reporting_Plugins) says to provide explicit version in `<reporting>`
> * the actual behavior of Maven itself is indeed to:
> 
> > Maven allows the reporting plugin version to be resolved from  when it is not specified in .
> 
> Can you link to some documentation that support:
> 
> > Maven allows the reporting plugin version to be resolved from  when it is not specified in
> 
> ?

Thanks for pointing this out. I’m still learning the Maven reporting/plugin resolution behavior, so I may be misunderstanding part of it.

I found the official Maven Model documentation for `ReportPlugin#getVersion()`, which states that starting with Maven 3, if no version is defined explicitly, Maven searches for the version in `build/plugins` and then `build/pluginManagement`:

https://maven.apache.org/ref/3.9.15/apidocs/org/apache/maven/model/ReportPlugin.html

The Maven Model reference documents the same behavior for the `<reporting><plugins><plugin><version>` element:

https://maven.apache.org/ref/3.9.15/maven-model/maven.html

So my current understanding is that the version can be resolved from `build/pluginManagement` even when it is not explicitly specified under `<reporting>`.

In the #916 reproducer, the report plugin is intentionally left without a version under `<reporting>`, while its version is managed under `<build><pluginManagement>`. The issue appears to be that `project.getReportArtifacts()` can still expose that report artifact with `RELEASE`, which is what causes the version comparison to differ from the corresponding build plugin.

I’ll review this behavior further to make sure I’m understanding Maven’s resolution correctly.

pzygielo:
> I found the official Maven Model documentation for `ReportPlugin#getVersion()`, which states that starting with Maven 3, if no version is defined explicitly, Maven searches for the version in `build/plugins` and then `build/pluginManagement`:
> 
> https://maven.apache.org/ref/3.9.15/apidocs/org/apache/maven/model/ReportPlugin.html
> 
> The Maven Model reference documents the same behavior for the `<reporting><plugins><plugin><version>` element:
> 
> https://maven.apache.org/ref/3.9.15/maven-model/maven.html
> 
> So my current understanding is that the version can be resolved from `build/pluginManagement` even when it is not explicitly specified under `<reporting>`.

Perfect! Thanks!

dhruv9b:
> Perfect! Thanks!

Thanks again for the question — I looked further into Maven's report plugin resolution to make sure the implementation matches Maven's behavior.

I found that the effective reporting plugin version is resolved in this order: an explicitly configured `<reporting>` version, then `<build><plugins>`, then `<build><pluginManagement>`.

My initial implementation only covered the explicit reporting version and `pluginManagement`. I’ve now updated the PR to also check `build/plugins` before `pluginManagement`, and added a unit test to verify that precedence.

I also verified the original #916 reproducer with the Maven Invoker integration test, which now passes successfully.

slawekjaranowski:
please preserve line:

```
 I hereby declare this contribution to be licenced under the Apache License Version 2.0, January 2004
```

form PR template in PR description

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/RequireSameVersions.java
```diff
@@ -136,6 +140,66 @@ private Map<String, List<String>> collectVersionMembers(
         return versionMembers;
     }
 
+    private Map<String, List<String>> collectReportVersionMembers(
```
slawekjaranowski, line 143: This is a copy of `collectVersionMembers` that differs in exactly one line (the version lookup at 160), and the copy also dropped the `// pattern is groupId[:artifactId[:type[:classifier]]]` comment. Please keep a single method and parameterise the version lookup instead, e.g.

```java
private Map<String, List<String>> collectVersionMembers(
        Set<Artifact> artifacts, Collection<String> patterns, String source,
        Function<Artifact, String> versionResolver) {
```

with the two existing call sites passing the current expression and the reporting one passing `this::getReportPluginVersion`.

enforcer-rules/src/main/java/org/apache/maven/enforcer/rules/RequireSameVersions.java
```diff
@@ -136,6 +140,66 @@ private Map<String, List<String>> collectVersionMembers(
...
+                    versionMembers
+                            .computeIfAbsent(version, unused -> new ArrayList<>())
+                            .add(artifact.getDependencyConflictId() + source);
+                }
+            }
+        }
+        return versionMembers;
+    }
+
+    private String getReportPluginVersion(Artifact artifact) {
```
slawekjaranowski, line 170: Worth considering a simpler shape here. Both `getReportArtifacts()` and `getReportPlugins()` are `@Deprecated` in `MavenProject`, and this adds a second use of that API. Since the reporting artifacts carry no usable version anyway, the entries could be built straight from the model (`project.getModel().getReporting()`), where `ReportPlugin` already has groupId/artifactId/version - that removes the artifact to model lookup entirely. The only thing to reconstruct would be `groupId:artifactId:maven-plugin` for the pattern matching and the message. Not a blocker, but it would make this special case considerably smaller.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/TestRequireSameVersions.java
```diff
@@ -254,6 +258,43 @@ void testProjectWithDifferentDependencyVersionsInBuildAndReportAndPlugins() thro
         assertThatCode(rule::execute).isInstanceOf(EnforcerRuleException.class);
     }
 
+    @Test
+    void testReportPluginUsesManagedVersionWhenArtifactVersionIsRelease() throws IOException {
```
slawekjaranowski, line 262: Both new tests only assert `doesNotThrowAnyException()`. Could you add a negative one - a genuine mismatch coming through `pluginManagement` (e.g. managed `maven-surefire-report-plugin` `3.6.0` against build `maven-surefire-plugin` `3.5.3`) that still has to fail? As it stands, an implementation that simply always returned the build plugin's version would pass both new tests, so nothing guards against the fix masking real mismatches.

enforcer-rules/src/test/java/org/apache/maven/enforcer/rules/TestRequireSameVersions.java
```diff
@@ -262,6 +303,51 @@ private static String extractGaString(Artifact dependency) {
         return String.format("%s:%s", dependency.getGroupId(), dependency.getArtifactId());
     }
 
+    @Test
+    void testReportPluginUsesBuildPluginVersionBeforeManagedVersion() throws IOException {
```
slawekjaranowski, line 307: This test sits after the private helpers `constructArtifact` / `extractGaString`, which breaks the file layout (tests first, helpers at the bottom). Please move it up next to the other tests.

maven-enforcer-plugin/src/it/projects/require-same-versions-916/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0
+                             http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.enforcer.its</groupId>
+  <artifactId>requireSameVersions-916</artifactId>
```
slawekjaranowski, line 29: Two nits on the IT. The directory name doesn't follow the convention of the neighbouring ITs (`require-same-versions_success`, `_failure`, `_multi-module-build`) - something like `require-same-versions_managed-report-plugin` would describe the case rather than the issue number. Also the license header uses `* ` prefixes inside the XML comment, unlike the other IT poms.

slawekjaranowski (commented):
I checked this out locally and verified the premise empirically, since that was the open question above.

Probe: a project with `<reporting>` declaring `maven-surefire-report-plugin` without a version, and `<build><pluginManagement>` pinning it to `3.2.5`. Running `mvn site` offline, so the resolution is visible:

```
Failed to get report for org.apache.maven.plugins:maven-surefire-report-plugin:
Plugin org.apache.maven.plugins:maven-surefire-report-plugin:3.2.5
or one of its dependencies could not be resolved
```

Maven really does use `3.2.5` from `<build><pluginManagement>`, not `RELEASE`. So to answer @pzygielo's question: the version genuinely is resolved from `pluginManagement`, `project.getReportArtifacts()` is what exposes the misleading `RELEASE`, and the rule was reporting a false mismatch. This matches the lookup order documented on `ReportPlugin#getVersion()` (Maven 3+: `build/plugins`, then `build/pluginManagement`), which is also the order implemented here.

The IT is a real regression test - I ran it both ways:

* `master` plus only the new IT directory: `require-same-versions-916/pom.xml ... FAILED`, the other 5 `require-same-versions` ITs pass
* this branch: all 6 pass

Unit tests pass too. So the direction is right; some comments on the implementation inline.

dhruv9b:
@slawekjaranowski Thanks for the review! I’ve addressed all the requested changes, including the consolidated helper, negative test for mismatched managed report-plugin versions, test/helper reorganization, and IT rename/license formatting.

The unit and integration tests are passing. Ready for another review.

slawekjaranowski (approved):

slawekjaranowski:
ok, it is god enough, deprecated methods can be removed next time 😄

PR #1025, opened 2026-09-13
[MENFORCER-517] RequireReleaseDeps checks now also for optional dependencies

### LINKED ISSUE #855: [MENFORCER-517] requireReleaseDeps ignores optioanal dependencies
**[James Nord](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=jnord_cbs)** opened **[MENFORCER-517](https://issues.apache.org/jira/browse/MENFORCER-517?redirect=false)** and commented

A project configured with an optional dependency with a snapshot version and enforcer with `requireReleaseDeps` set to fail should fail the build, however the rule does not look at optional dependencies.

Whilst this makes not difference to consumers, it means that  `requireReleaseDeps` can not be used to help enforce the reproducibility of release builds.

### Steps to Reproduce

create a maven pom with the following: 

```java
<?xml version="1.0" encoding="utf-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>  <groupId>maven.test</groupId>
  <artifactId>enforcer-bug</artifactId>
  <packaging>jar</packaging>
  <version>1-SNAPSHOT</version>  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <version>3.5.0</version>
        </plugin>
      </plugins>
    </pluginManagement>    <plugins>
      <plugin>
        <executions>
          <execution>
            <id>enforce-no-snapshots</id>
            <goals>
              <goal>enforce</goal>
            </goals>
            <configuration>
              <rules>
                <requireReleaseDeps>
                  <message>No Snapshots Allowed For Release Versions</message>
                  <onlyWhenRelease>false</onlyWhenRelease>
                </requireReleaseDeps>
              </rules>
              <fail>true</fail>
              <skip>false</skip>
            </configuration>
          </execution>
        </executions>
      </plugin>
    </plugins>
  </build>  <dependencies>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.1-SNAPSHOT</version>
      <type>jar</type>
      <scope>test</scope>
      <optional>true</optional>
    </dependency>
  </dependencies></project>
 
```

run `mvn validate`

### Expected results

The build fails with

```java
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-enforcer-plugin:3.5.0:enforce (enforce-no-snapshots) on project enforcer-bug:
[ERROR] Rule 0: org.apache.maven.enforcer.rules.dependency.RequireReleaseDeps failed with message:
[ERROR] No Snapshots Allowed For Release Versions
[ERROR] maven.test:enforcer-bug:jar:1-SNAPSHOT
[ERROR]    junit:junit:jar:4.1-SNAPSHOT <--- is not a release dependency 
```

### Actual results

the build passes

```java
[INFO] Rule 0: org.apache.maven.enforcer.rules.dependency.RequireReleaseDeps passed 
```

Note if you remove the `<optional>true</optional>` from the dependency you will see that the enforcer rule passes

---

**Affects:** 3.5.0

schwarfl:
This pull-request fixes issue https://github.com/apache/maven-enforcer/issues/855 by checking also optional dependencies for snapshot versions.

- [X] Your pull request should address just one issue, without pulling in other changes.
- [X ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [X] Run `mvn verify` to make sure basic checks pass.
- [X] You have run the integration tests successfully (`mvn -Prun-its verify`).

- [ X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)


slawekjaranowski:
Thanks for digging into this!

Before we go further with the current approach — I think the root cause sits one level deeper, and this is actually a regression.

`BannedDependenciesBase#execute()` resolves the tree with:

```java
DependencyNode rootNode = resolverUtil.resolveTransitiveDependenciesVerbose(Collections.emptyList());
```

and `ResolverUtil#resolveTransitiveDependenciesVerbose(List)` hardcodes `excludeOptional = true`, so optional dependencies are filtered out of the `CollectRequest` before the tree is even built. The rule never gets a chance to see them.

That filtering was introduced in https://github.com/apache/maven-enforcer/pull/256 (commit 5e0ca5a, first released in 3.3.0). Before that change `BannedDependenciesBase` called `resolveTransitiveDependenciesVerbose()` with no selectors, i.e. with the session defaults — and the default `OptionalDependencySelector` only drops optional dependencies from depth 2 on:

```java
public boolean selectDependency(Dependency dependency) {
    return depth < 2 || !dependency.isOptional();
}
```

so direct optional dependencies were included. `DependencyConvergence` and `RequireUpperBoundDeps` previously used `AllLevelsOptionalDependencySelector`, so for them `excludeOptional = true` is correct and should stay — only `BannedDependenciesBase` lost behaviour it used to have.

So the fix should live in `BannedDependenciesBase`, not in `RequireReleaseDeps`:

```java
DependencyNode rootNode = resolverUtil.resolveTransitiveDependencies(true, false, Collections.emptyList());
```

(or, for readability, a new `resolveTransitiveDependenciesVerbose(boolean excludeOptional, List<String> excludedScopes)` overload in `ResolverUtil`).

Advantages over the current patch:

- optional dependencies go through the normal tree walk, so the failure message keeps the usual tree format:
  ```
  org.apache.maven.enforcer.its:requireReleaseDepsOptional:jar:1.0-SNAPSHOT
     org.apache.maven.plugins.enforcer.its:menforcer85_api:jar:1.0-SNAPSHOT <--- is not a release dependency
  ```
  which is exactly what the reporter expected in #855
- transitive dependencies *of* an optional dependency are covered too, not only the direct ones
- `includes`/`excludes` keep working through the existing code path
- `bannedDependencies` — same base class, same blind spot — is fixed at the same time
- no manual `DefaultArtifact` construction, so no need to reimplement the `ArtifactTypeRegistry` mapping (`test-jar` → classifier `tests`, `pom`/`bundle` extensions, …)

I tried it locally: your IT `require-release-dependencies-optional_failure` passes with that one-line change, and `require-release-dependencies*`, `multimodule-require-release*`, `banned-dependencies*` and `ban-transitive*` all stay green. The only thing that needs touching is the Mockito stubs in `RequireReleaseDepsTest` and `BannedDependenciesTest`, which currently stub `resolveTransitiveDependenciesVerbose(anyList())`.

Could you rework the PR along these lines? Please also:

- prefix the commit message and PR title with `[MENFORCER-517]`
- rebase on current `master`

slawekjaranowski (changes requested):
.

schwarfl:
Hi, 
you are right, your solution is much better. I added the changes you requested

PR #1035, opened 2026-09-20
Add the evaluateGroovy rule

slachiewicz:
Groovy-backed counterpart of `evaluateBeanshell`, which is deprecated ahead of the BeanShell retirement (apache/maven#13207). Same shape: `condition` (Maven `${…}` interpolated first, then evaluated by `GroovyShell`, must yield a boolean) and `message`.

Groovy is an `optional` dependency of enforcer-rules rather than a hard one: it is about 7 MB and the enforcer runs in nearly every build. Users add `org.apache.groovy:groovy` to the plugin declaration (shown on the rule page); without it the rule fails with a one-line message saying exactly that. The Groovy call sits in a nested class so the rule itself loads and can report the absence.

The rule page notes the one porting gotcha: BeanShell compared bare words (`${project.artifactId} == foo`), Groovy needs both sides quoted. The deprecation notices on `evaluateBeanshell` now point here.

Verified: `TestEvaluateGroovy` 8/8; ITs `evaluate-groovy` (two rules pass), `evaluate-groovy-missing` (expected failure with the classpath message) and `evaluate-beanshell` → Passed: 3; `apache-rat:check` clean; `mvn site -pl enforcer-rules` renders the new page.


enforcer-rules/src/site/markdown/evaluateGroovy.md.vm
```diff
@@ -0,0 +1,86 @@
+---
+title: Beanshell
+author: 
+  - Brian Fox
```
slawekjaranowski: ?

enforcer-rules/src/site/markdown/evaluateGroovy.md.vm
```diff
@@ -0,0 +1,86 @@
+---
+title: Beanshell
```
slawekjaranowski: ?

enforcer-rules/pom.xml
```diff
@@ -112,6 +112,13 @@
       <artifactId>bsh</artifactId>
       <version>2.0b6</version>
     </dependency>
+    <dependency>
+      <!-- for evaluateGroovy; not shipped with the plugin, users add it to the plugin declaration -->
```
slawekjaranowski: To check if optional is not added to plugin dependencies, maybe scope provided

slawekjaranowski (approved):

